PR #7, opened 2019-02-06
New API for dumping current configuration

Pols95:


carapace-server/src/main/java/org/carapaceproxy/api/ConfigResource.java
```diff
@@ -47,10 +54,40 @@
 public class ConfigResource {
 
     private static final Logger LOG = Logger.getLogger(ConfigResource.class.getName());
+    private static final String CONFIGURATION_DUMPING_FILENAME = "server.dynamic.config";
 
     @javax.ws.rs.core.Context
     ServletContext context;
 
+    @GET
```
eolivelli: add @Produces("text/plain")

carapace-server/src/main/java/org/carapaceproxy/api/ConfigResource.java
```diff
@@ -47,10 +54,40 @@
...
+    public Response dump() {
+        StreamingOutput configStream = (output) -> {
```
eolivelli: the result is not so big, we can simply use a String and not the StreamingOutput.
I would be fine to use StreamingOutput but you have not to create the String.

Having the String allows the server to send the Content-Length header and this way it is clear to understand if the file has been downloaded completely

carapace-server/src/main/java/org/carapaceproxy/api/ConfigResource.java
```diff
@@ -47,10 +54,40 @@
...
+                });
+                byte[] data = builder.toString().getBytes("UTF-8");
+                output.write(data);
+                output.flush();
+            } catch (Exception e) {
+                throw new WebApplicationException("Unable to dump current configuration.");
+            }
+        };
+        LOG.info("Dumped current configuration from API");
+        return Response.ok(configStream, MediaType.APPLICATION_OCTET_STREAM)
```
eolivelli: "text/plain" is consistent with the other API methods in this file.
You should simply add the annotation as suggested above and then let this method return a simple java.lang.String

carapace-server/src/main/java/org/carapaceproxy/api/ConfigResource.java
```diff
@@ -47,10 +54,40 @@
...
     @javax.ws.rs.core.Context
     ServletContext context;
 
+    @GET
+    public Response dump() {
+        StreamingOutput configStream = (output) -> {
+            try {
+                HttpProxyServer server = (HttpProxyServer) context.getAttribute("server");
+                ConfigurationStore configStore = server.getDynamicConfigurationStore();
+                String lineSeparator = configStore.getProperty("line.separator", "\n");
```
eolivelli: this shouldn't be needed
we can use LF (unix style)

another approach is to use the native platform line separator (System.getProperty("line.separator")) 
https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
```diff
@@ -167,6 +167,7 @@ public void commitConfiguration(ConfigurationStore newConfigurationStore) {
                             psInsert.setString(2, v);
                             psInsert.executeUpdate();
                         }
+                        properties.put(k, v);
```
eolivelli: we should apply this change only if "con.commit()" completes successfully

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
```diff
@@ -176,6 +177,7 @@ public void commitConfiguration(ConfigurationStore newConfigurationStore) {
                         LOG.log(Level.INFO, "Deleting '" + k + "'");
                         psDelete.setString(1, k);
                         psDelete.executeUpdate();
+                        properties.remove(k);
```
eolivelli: we should apply this change only if "con.commit()" completes successfully

.gitignore
```diff
@@ -8,3 +8,4 @@
 dependency-reduced-pom.xml
 **/nbproject/
 **/access.log
+carapace-server/dbdata/*
```
eolivelli: Is this needed?

Pols95: With this commit tests about persistence on herddb now uses a temp dir, so it's useless indeed. I'll remove it.

PR #14, opened 2019-02-09
Simple rest api cli

Pols95:


eolivelli:
Please rebase onto current master

api-cli.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+if [ "$#" -ne 2 ]
+then
+    echo "Usage: dump|validate|apply configuration.config"
+    exit
+fi
+case "$1" in
+    "dump")
+        curl -X GET localhost:8001/api/config --user admin:admin -o "$2"
```
eolivelli: we should make localhost:8081 a parameter (with default localhost:8081)
we should also have a parameter for "username" and "password"

I think that 'dump' by default should write to stdout (without '-o')
we can have a parameter for the destination file as well, with default to stdout

it will be very easy to use stdout 

`api-cli.sh dump --user myuser --password  > myfile.conf`

Pols95:
> can we change the name to "carapace-cli.sh" ?
> 
> I think that as a short term solution the bash script works
> but we will soon need something more complex and I guess we will switch to a Java program.
> 
> It will be easy to create our cli using JCommander and JAX-RS.
> If you want we can clean up this patch and merge, then start thinking about JCommander
> 
> how does this sounds to you ?
> 
> thanks

Maybe the best solution is to erase this one and restarting with jcommander

eolivelli:
@Pols95 Yes, better to work on a Java CLI. thanks

Issue #10, opened 2019-02-08
Enable the user to see the current configuration from the UI


eolivelli:
Add views on the UI to see:
- configured backends (yet done)
- configured listeners (yet done)
- configured routes
- configured actions
- configured directors

A full dump of current configuration

Pols95:
> Add views on the UI to see:
> 
> * configured actions
> * configured backends
> * configured routes
> 
> A full dump of current configuration

Pols95:
Maybe the best way is to add a text-area to see and directly update the configuration?

eolivelli:
IMO we must have the textarea
But we should also have some user friendly way to see what's really going on, that is the effective configuration that the system is using

Pols95:
Disabled backends are ignored by endpoint mapper, so it's correct that are hidden in the UI (current behavior)

PR #16, opened 2019-02-10
Enable the user to see the current configuration from the UI

Pols95:
This is just the first step of enabling user to see current config (#10 ).
What about merging what done up to now (so we can go ahead testing easier) and then develop later (asap) the remaining parts? @eolivelli

### DIFF
--- a/carapace-ui/src/main/webapp/public/index.html
+++ b/carapace-ui/src/main/webapp/public/index.html
@@ -5,7 +5,6 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <title>Http Proxy UI</title>
-    <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
   </head>
   <body>
     <noscript>
--- a/carapace-ui/src/main/webapp/public/index.html
+++ b/carapace-ui/src/main/webapp/public/index.html
@@ -5,6 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <title>Http Proxy UI</title>
+    <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
   </head>
   <body>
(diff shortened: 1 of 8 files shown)


carapace-ui/src/main/webapp/public/index.html
```diff
@@ -5,6 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <title>Http Proxy UI</title>
+    <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
```
eolivelli: Better not to use cdn but copy the contents, please check locense

carapace-ui/src/main/webapp/src/components/Configuration.vue
```diff
@@ -0,0 +1,82 @@
...
+            this.reload()
+        },
+        methods: {
+            save: function () {
+                var self = this
+                axios.post('/api/config/apply', this.configuration, this.axiosParams).then(function (response) {
+                    var data = response.data
+                    self.opSuccess = data.ok
+                    self.opMessage = data.ok ? "Configuration saved successfully." : "Error on configuration saving: " + data.error
+                    console.log(response)
```
eolivelli: Remove debug

carapace-ui/src/main/webapp/src/components/Configuration.vue
```diff
@@ -0,0 +1,82 @@
...
+                    self.opSuccess = false
+                    self.opMessage = "Error on reloading current configuration:" + error
+                    console.log(error)
+                })
+            }
+        }
+    }
+</script>
+
+<style>
```
eolivelli: Is this the right place? Don't we have some common place for css?

Pols95: I changed it as component-scoped style

eolivelli:
@nicoloboschi  please take another look

nicoloboschi:
Look good to me (we should rename mockserver javascript file but we can do later)

PR #26, opened 2019-02-19
UI: Show dynamic certificate status

Pols95:


carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
```diff
@@ -62,6 +74,18 @@ public String getHostname() {
             return hostname;
         }
 
+        public boolean isDynamic() {
+            return dynamic;
+        }
+
+        public DynamicCertificateState getStatus() {
```
eolivelli: Do not expose business logic beans on the API.
Please create a bean (inner static public class here:

Pols95: Maybe for the frontend it could be simply a string

eolivelli: Sure

eolivelli: I meant something different.
If you call toString() the API would break if someone adds/remove values and/or changes the toString() method.
You have to explicitly transcode the values. a String is fine but not this way

PR #28, opened 2019-02-20
Enable the user to see the current configuration from the UI

Pols95:
Views (+ REST API requests) integrated:
- Routes
- Actions
- Directors
- Metrics

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/api/MetricsResource.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/api/MetricsResource.java
@@ -17,16 +17,32 @@
  under the License.
 
  */
-package org.carapaceproxy.server.config;
+package org.carapaceproxy.api;
 
-import io.netty.handler.codec.http.HttpRequest;
+import javax.servlet.ServletContext;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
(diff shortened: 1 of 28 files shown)


carapace-server/src/main/java/org/carapaceproxy/api/MetricsResource.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+@Path("/metrics")
+@Produces("application/json")
+public class MetricsResource {
+
+    @javax.ws.rs.core.Context
+    ServletContext context;
+
+    @GET
+    @Path("path")
```
eolivelli: 'url' ?

carapace-server/src/main/java/org/carapaceproxy/server/config/RequestMatcher.java
```diff
@@ -27,4 +27,6 @@
 public interface RequestMatcher {
 
     RoutingKey matches(HttpRequest request);
+
+    String appliedPattern();
```
eolivelli: I think this is not applicable for every 'RequestMatcher', think about, we will have 'AndRequestMatcher', 'OrRequestMatcher', 'NegateRequestMatcher'

I suggest you to have a 'getDescription()' with a javadoc which explains that this is a description for the UI
this way every Matcher will create a suitable message

PR #32, opened 2019-02-22
Let's the BackendHeathManager period to be configured at runtime

Pols95:


carapace-server/src/main/java/org/carapaceproxy/server/backends/BackendHealthManager.java
```diff
@@ -201,16 +204,30 @@ public boolean isAvailable(String id) {
...
             LOG.info("Applying new connect timeout " + this.connectTimeout + " ms");
         }
+
         this.mapper = mapper;
+
+        this.period.set(newPeriod);
+        LOG.info("Applying health probe period " + newPeriod + " ms");
+
+        if (timer != null) {
+            start();
```
eolivelli: We can do better.
Just cancel the ScheduledFuture returned by scheduleAtfixedRate and schudule a new execution

carapace-server/src/main/java/org/carapaceproxy/server/backends/BackendHealthManager.java
```diff
@@ -65,26 +68,64 @@ public BackendHealthManager(RuntimeServerConfiguration conf, EndpointMapper mapp
...
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    public synchronized void reloadConfiguration(RuntimeServerConfiguration newConfiguration, EndpointMapper mapper) {
+        int newPeriod = newConfiguration.getHealthProbePeriod();
+
+        if (scheduledFuture != null) {
+            scheduledFuture.cancel(true);
```
eolivelli: We should cancel the timer only if period is changing

carapace-server/src/main/java/org/carapaceproxy/server/backends/BackendHealthManager.java
```diff
@@ -50,11 +52,12 @@
     private final StatsLogger mainLogger;
 
     private ScheduledExecutorService timer;
-    // configured only at start
-    private int period;
+    private ScheduledFuture<?> scheduledFuture;
 
     // can change at runtime
-    private volatile int connectTimeout = 60000;
+    private volatile AtomicInteger period = new AtomicInteger(0);
```
eolivelli: volatile and AtomicInteger does not make sense together.

We don't need AtomicInteger, we aren't performin compare and set operations and every access to period is inside a synchronized block

carapace-server/src/main/java/org/carapaceproxy/server/backends/BackendHealthManager.java
```diff
@@ -50,11 +52,12 @@
...
+    // can change at runtime
+    private volatile AtomicInteger connectTimeout = new AtomicInteger(60000);
```
eolivelli: Same here

eolivelli:
Please merge as soon as CI is green

PR #39, opened 2019-02-27
UI: improve Last probe result message

Pols95:


carapace-server/src/main/java/org/carapaceproxy/server/backends/BackendHealthCheck.java
```diff
@@ -95,56 +96,54 @@ public static BackendHealthCheck check(String host, int port, String path, int t
...
+
                 try (InputStream is = httpConn.getInputStream();) {
                     int httpCode = httpConn.getResponseCode();
-                    String httpBody = IOUtils.toString(is, StandardCharsets.UTF_8);
-                    if (httpCode >= 200 && httpCode <= 299) {
-                        return new BackendHealthCheck(startts, System.currentTimeMillis(), RESULT_SUCCESS, httpBody);
-                    } else {
-                        return new BackendHealthCheck(
-                            startts, System.currentTimeMillis(), RESULT_FAILURE_STATUS, 
-                                "HttpCode="+httpCode+", HttpMessage="+httpConn.getResponseMessage()+", HttpBody="+httpBody);
```
eolivelli: Why are you dropping the body? It is very usefull

Issue #42, opened 2019-03-20
Add ZooKeeper Watch on configuration changes


Pols95:
In a clustered setup when an user modifies the configuration all of the other peers have to apply the new configuration as well.

Subtasks:
- create a /configurationstatus z-node
- add a watcher on that node
- when configuration is changed the server which committed the configuration will perform a write to that z-node, and so other peers will reload configuration

PR #47, opened 2019-03-26
Introduce ZooKeeper based group membership

eolivelli:
Introduce GroupMembershipHandler
Introduce first implementation with ZooKeeper
Introduce mode=cluster|standalone

relates to #42

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandler.java
@@ -29,6 +29,7 @@ import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.CuratorWatcher;
 import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.WatchedEvent;
 import org.apache.zookeeper.data.Stat;
 import org.carapaceproxy.cluster.GroupMembershipHandler;
@@ -63,6 +64,7 @@ public class ZooKeeperGroupMembershipHandler implements GroupMembershipHandler, 
             if (exists == null) {
                 client.create()
                         .creatingParentsIfNeeded()
+                        .withMode(CreateMode.EPHEMERAL) // auto delete on close
                         .forPath("/proxy/peers/" + peerID);
             }
         } catch (Exception ex) {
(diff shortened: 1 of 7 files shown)


eolivelli:
@Pols95 @diegosalvi  this is this work-in-progress but is is open for early review

carapace-server/src/main/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandler.java
```diff
@@ -0,0 +1,167 @@
...
+    public void watchEvent(String eventId, EventCallback callback) {
+        try {
+            final String path = "/proxy/events";
+
+            LOG.info("watching " + path);
+            PathChildrenCache cache = new PathChildrenCache(client, path, true);
+            // hold a strong reference to the PathChildrenCache
+            watchedEvents.add(cache);
+            cache.getListenable().addListener((PathChildrenCacheListener) (CuratorFramework cf, PathChildrenCacheEvent pcce) -> {
+                LOG.log(Level.INFO, "ZK event {0} at {1}", new Object[]{pcce, path});
```
eolivelli: this is still work in progress.
watchers are getting notified about every event

eolivelli:
@diegosalvi  @Pols95  can you please take a look and eventually merge ?
This way I can move forward.

mino181295:
Looking at the pullrequest I noticed only some improvements/opinions 

- Nodes paths are hardcoded. Maybe would be better to store them in constants also for testing what really happens in the TestingServer.
- Is it really necessary to pass a "mode" ? I think `zkAddress` is enough to enable cluster.
- There is no way to start the server via script doing like `bin/service zookeeper start` actually
- In the provided conf/ example file maybe should be provided a commented example of "mode", "zkAddress", zkTimeout
- Instead of using `Thread.sleep` could be implemented a trivial waitForCondition utility

For me it's okay, maybe these suggestions could be implementend in further commits

eolivelli:
> Looking at the pullrequest I noticed only some improvements/opinions
> 
> * Nodes paths are hardcoded. Maybe would be better to store them in constants also for testing what really happens in the TestingServer.

I will make the path configurable but we don't expect to share the zk cluster with other applications

> * Is it really necessary to pass a "mode" ? I think `zkAddress` is enough to enable cluster.
It is only the first step, with that flag we will start other components like the database in 'replicated'. That's the content of the follow up patch
> * There is no way to start the server via script doing like `bin/service zookeeper start` actually
It will be a good idea, but it is only for test/demo because it is better to havehave a real zk cluster in production
> * In the provided conf/ example file maybe should be provided a commented example of "mode", "zkAddress", zkTimeout
As last step we will provide complete examples and dcos
> * Instead of using `Thread.sleep` could be implemented a trivial waitForCondition utility
We don't have waitForCondition yet, feel free to send a patch, I would prefer to keep it out of this patch
> 
> For me it's okay, maybe these suggestions could be implementend in further commits

Thanks

Issue #34, opened 2019-02-23
Set Carapace to use db configuration as default


pv3ntur1:
Simply we need to set set in server.properties
config.type=database

And then in the HerdDBConfigurationStore constructor call __commitConfiguration__
right after __loadCurrentConfiguration__, in order to store the properties in the file server.properties (whether the db was empty) as fallback configuration.

eolivelli:
@Pols95 I am not sure about the second part.
IMO no code change is needed.
The server will start with an empty database and it will remain in such state until the administration pushes a configuration from the API/UI

Pols95:
Maybe the simplest solution is to use the db always and let the admin to specify an optional config file (like it's actually done) that will be load to the db.

PR #53, opened 2019-04-19
Set Carapace to use db configuration as default #34

Pols95:


carapace-server/src/main/resources/bin/service
```diff
@@ -113,7 +116,7 @@ start() {
     fi
   fi
 
-  $PRERUN "$BASE_DIR/bin/java-utils.sh -daemon $SERVICE -Dpidfile=$PIDFILENAME $MAINCLASS $OTHERPARAMS"
+  $PRERUN "$BASE_DIR/bin/java-utils.sh -daemon $SERVICE -Dpidfile=$PIDFILENAME $MAINCLASS $OTHERPARAMS $DB_CONFIG_FILE"
```
eolivelli: It is better to let the service pick the default file server.properties
and not override it here.

with your change you can't run
`bin/service server start conf/mydemoconfig.properties`

this is very common across several projects

carapace-server/src/main/resources/bin/service
```diff
@@ -95,6 +95,9 @@ fi
 ### Set working directory
 cd "$BASE_DIR"
 
+### Set db configuration
+DB_CONFIG_FILE="$BASE_DIR/conf/server.db.properties"
```
eolivelli: this is to be removed

PR #58, opened 2019-04-24
Feature/#44

Pols95:
Now, in the configuration can be setup **mapper.debuggingheader.enable=true|false** (default false) to add the custom header **Routing-Path: Route-id: id_selected_route; Action-id: id_selected_action; Director-id: id_selected_director; Backend-id: id_selected_backend**


eolivelli:
Please rebase

PR #59, opened 2019-04-24
Option to add a debug header to see which route/action/director/backe…

Pols95:
Now, in the configuration can be setup **mapper.debug=true|false** (default false) and **mapper.debug.name=header-name** (default Routing-Path) to add a custom header **X-header-name: id_selected_route;id_selected_action;id_selected_director;id_selected_backend** to HttpResponses for actions of type *cache* and *proxy*

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
@@ -32,6 +32,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.carapaceproxy.EndpointMapper;
 import org.carapaceproxy.MapResult;
+import org.carapaceproxy.MapResult.Action;
 import org.carapaceproxy.configstore.ConfigurationStore;
 import org.carapaceproxy.server.RequestHandler;
 import static org.carapaceproxy.server.StaticContentsManager.DEFAULT_INTERNAL_SERVER_ERROR;
@@ -75,6 +76,9 @@ public class StandardEndpointMapper extends EndpointMapper {
     private static final String ACME_CHALLENGE_URI_PATTERN = "/\\.well-known/acme-challenge/";
     private DynamicCertificatesManager dynamicCertificateManger;
 
+    public static final String DEBUGGING_HEADER_ROUTING_PATH = "Routing-Path";
+    private boolean debuggingHeaderEnabled = false;
+
     public StandardEndpointMapper(BackendSelector backendSelector) {
         this.backendSelector = backendSelector;
     }
@@ -99,6 +103,8 @@ public class StandardEndpointMapper extends EndpointMapper {
         LOG.info("configured mapper.forcedirector.parameter=" + forceDirectorParameter);
         this.forceBackendParameter = properties.getProperty("mapper.forcebackend.parameter", forceBackendParameter);
         LOG.info("configured mapper.forcebackend.parameter=" + forceBackendParameter);
+        this.debuggingHeaderEnabled = Boolean.parseBoolean(properties.getProperty("mapper.debuggingheader.enable", "false"));
+        LOG.info("configured mapper.debuggingheader.enable=" + debuggingHeaderEnabled);
 
         for (int i = 0; i < MAX_IDS; i++) {
             String prefix = "header." + i + ".";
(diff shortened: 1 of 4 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -367,37 +373,38 @@ public MapResult map(HttpRequest request, String userId, String sessionId, Backe
...
                     }
+
+                    BackendConfiguration backend = this.backends.get(backendId);
+                    if (backend != null && backendHealthManager.isAvailable(backendId)) {
+                        if (this.debuggingHeaderEnabled) {
+                            String routingPath = "Route-id: " + route.getId() + "; ";
+                            routingPath += "Action-id: " + action.getId() + "; ";
+                            routingPath += "Director-id: " + action.getDirector() + "; ";
+                            routingPath += "Backend-id: " + backendId;
+                            action.addCustomHeader(new CustomHeader(DEBUGGING_HEADER_ROUTING_PATH, routingPath, HeaderMode.HEADER_MODE_ADD));
```
eolivelli: this is not good, "Actions" are immutable objects, you cannot add a "custom header"

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -99,6 +103,8 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
         LOG.info("configured mapper.forcedirector.parameter=" + forceDirectorParameter);
         this.forceBackendParameter = properties.getProperty("mapper.forcebackend.parameter", forceBackendParameter);
         LOG.info("configured mapper.forcebackend.parameter=" + forceBackendParameter);
+        this.debuggingHeaderEnabled = Boolean.parseBoolean(properties.getProperty("mapper.debuggingheader.enable", "false"));
```
eolivelli: what about "mapper.debug" ?

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -75,6 +76,9 @@
     private static final String ACME_CHALLENGE_URI_PATTERN = "/\\.well-known/acme-challenge/";
     private DynamicCertificatesManager dynamicCertificateManger;
 
+    public static final String DEBUGGING_HEADER_ROUTING_PATH = "Routing-Path";
```
eolivelli: better to make it configurable
the debug MUST be a name which starts with "X-", which means it is a custom header

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -367,37 +373,38 @@ public MapResult map(HttpRequest request, String userId, String sessionId, Backe
...
                             break;
                         default:
                             return MapResult.NOT_FOUND(route.getId());
                     }
+
+                    BackendConfiguration backend = this.backends.get(backendId);
+                    if (backend != null && backendHealthManager.isAvailable(backendId)) {
+                        if (this.debuggingHeaderEnabled) {
+                            String routingPath = "Route-id: " + route.getId() + "; ";
+                            routingPath += "Action-id: " + action.getId() + "; ";
```
eolivelli: this is too verbose, just "id;id;id;id" is enough

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
Pols95: Do you think it's ok cloning the object ('cause i think it's good adding all additional headers at this higher level as CustomHeader)? Or is it better to add it in the request handler (lower level)?

eolivelli: Clone = allocate...
If we can save allocations on the hot path it is better.
So I prefer not to clone

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -99,6 +104,11 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
         LOG.info("configured mapper.forcedirector.parameter=" + forceDirectorParameter);
         this.forceBackendParameter = properties.getProperty("mapper.forcebackend.parameter", forceBackendParameter);
         LOG.info("configured mapper.forcebackend.parameter=" + forceBackendParameter);
+        // To add custom debugging header for request choosen mapping-path
+        this.debuggingHeaderEnabled = Boolean.parseBoolean(properties.getProperty("mapper.debug", "false"));
+        LOG.info("configured mapper.debug=" + debuggingHeaderEnabled);
+        this.debuggingHeaderName = "X-" + properties.getProperty("mapper.debug.name", DEBUGGING_HEADER_DEFAULT_NAME);
```
eolivelli: this way you are forcing that prefix even on the user-configured value.
It is better to have a default of "X-Proxy-Path" and let the user override the value

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -367,37 +377,39 @@ public MapResult map(HttpRequest request, String userId, String sessionId, Backe
...
+                            selectedAction = MapResult.Action.CACHE;
                             break;
                         default:
                             return MapResult.NOT_FOUND(route.getId());
                     }
+
+                    BackendConfiguration backend = this.backends.get(backendId);
+                    if (backend != null && backendHealthManager.isAvailable(backendId)) {
+                        List<CustomHeader> customHeaders = new ArrayList(action.getCustomHeaders());
+                        if (this.debuggingHeaderEnabled) {
```
eolivelli: if this debug is not enabled you can skip the clone of the arraylist.

please also ensure that action.getCustomHeaders() returns an unmodifiableList

Issue #43, opened 2019-03-21
route match show  regex with back-slash  configuration 


hamadodene:
example: 
` "test\.test\.test\/test/.*" `
if i set this string,  after reload  the regex change in --> test.test.test/test/.* 
if i save my current configuration carapace override  `"test\.test\.test\/test/.*"` with  test.test.test/test/.* .

Pols95:
It's a db saving bug... Regex patterns are saved as string with "\" yet processed

PR #64, opened 2019-04-30
route match show regex with back-slash configuration #43

Pols95:


carapace-server/src/main/java/org/carapaceproxy/api/ConfigResource.java
```diff
@@ -96,7 +96,7 @@ private PropertiesConfigurationStore buildStore(String newConfiguration) throws
         }
         Properties properties = new Properties();
         try {
-            properties.load(new StringReader(newConfiguration));
+            properties.load(new StringReader(newConfiguration.replace("\\", "\\\\"))); // to preserve backslashes
```
eolivelli: we should add a test !

Pols95:
@eolivelli 'cause configuration is managed by a java properties file, backslashes '\' has to be explicitly escaped in order to preserve it from Properties#load (https://docs.oracle.com/javase/6/docs/api/java/util/Properties.html#load(java.io.Reader))

PR #65, opened 2019-05-01
UI: allow to see requests issued to the proxy (up to now only backend…

Pols95:
Now for each listener can be seen total requests count.


carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
```diff
@@ -132,6 +136,7 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) {
                 currentRequest.clientRequestFinished(trailer);
                 totalRequests.inc();
                 runningRequests.inc();
+                listenerRequests.inc();
```
eolivelli: Is there a difference from totalRequests?

Pols95: None, the sum of each listener requests is propertly equal to the total amount.

PR #66, opened 2019-05-03
48 Ability to force https connections on given URIs: error and/or red…

Pols95:
**New Redirect Action:**
```
action.1.type=redirect    # new type of action
action.1.code=307    # http redirect code
action.1.redirect.location=https://localhost:1234/index.html    # for absolute/relative redirection s
action.1.redirect.proto=https    # 'https' for redirect to https (default http is used)
action.1.redirect.host=192.0.0.1    # redirection host
action.1.redirect.port=1234    # redirection port
action.1.redirect.path=/indexX.html    # redirection path
```
Two way to use it:
* with redirec.location: to specify realive/absolute redirections
* with redirect.proto/host/port/path: to customize a particular part of the redirection path (header Location)

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
@@ -63,6 +63,8 @@ public class ClientConnectionHandler extends SimpleChannelInboundHandler<Object>
     volatile boolean refuseOtherRequests;
     private final List<RequestHandler> pendingRequests = new CopyOnWriteArrayList<>();
     final Runnable onClientDisconnected;
+    private final String listenerHost;
+    private final int listenerPort;
 
     public ClientConnectionHandler(
             StatsLogger mainLogger,
@@ -91,10 +93,20 @@ public class ClientConnectionHandler extends SimpleChannelInboundHandler<Object>
         this.onClientDisconnected = onClientDisconnected;
         this.backendHealthManager = backendHealthManager;
         this.requestsLogger = requestsLogger;
+        this.listenerHost = listenerHost;
+        this.listenerPort = listenerPort;
     }
 
     public SocketAddress getClientAddress() {
         return clientAddress;
+    }
+
+    public String getListenerHost() {
+        return listenerHost;
+    }
+
+    public int getListenerPort() {
+        return listenerPort;
     }
 
     @Override
--- a/carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
@@ -62,9 +62,7 @@ public class ClientConnectionHandler extends SimpleChannelInboundHandler<Object>
     volatile Boolean keepAlive;
     volatile boolean refuseOtherRequests;
     private final List<RequestHandler> pendingRequests = new CopyOnWriteArrayList<>();
-    final Runnable onClientDisconnected;
-    private final String listenerHost;
-    private final int listenerPort;
+    final Runnable onClientDisconnected;  
 
     public ClientConnectionHandler(
             StatsLogger mainLogger,
@@ -92,21 +90,11 @@ public class ClientConnectionHandler extends SimpleChannelInboundHandler<Object>
         this.connectionStartsTs = System.nanoTime();
         this.onClientDisconnected = onClientDisconnected;
         this.backendHealthManager = backendHealthManager;
-        this.requestsLogger = requestsLogger;
-        this.listenerHost = listenerHost;
-        this.listenerPort = listenerPort;
+        this.requestsLogger = requestsLogger; 
     }
 
     public SocketAddress getClientAddress() {
         return clientAddress;
-    }
-
-    public String getListenerHost() {
-        return listenerHost;
-    }
-
-    public int getListenerPort() {
-        return listenerPort;
     }
 
     @Override
--- a/carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
@@ -58,12 +58,12 @@ import org.carapaceproxy.client.EndpointConnection;
 import org.carapaceproxy.client.EndpointKey;
 import org.carapaceproxy.client.EndpointNotAvailableException;
 import org.carapaceproxy.client.impl.EndpointConnectionImpl;
-import static org.carapaceproxy.server.StaticContentsManager.DEFAULT_INTERNAL_SERVER_ERROR;
 import org.carapaceproxy.server.backends.BackendHealthManager;
 import org.carapaceproxy.server.cache.ContentsCache;
 import org.carapaceproxy.server.filters.UrlEncodedQueryString;
 import org.apache.bookkeeper.stats.Counter;
 import org.apache.bookkeeper.stats.StatsLogger;
+import static org.carapaceproxy.server.StaticContentsManager.DEFAULT_INTERNAL_SERVER_ERROR;
 import static org.carapaceproxy.server.mapper.CustomHeader.HeaderMode.HEADER_MODE_ADD;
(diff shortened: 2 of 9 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -386,6 +392,42 @@ private void serveStaticMessage() {
...
+        String host = this.connectionToClient.getListenerHost();
+        int port = this.connectionToClient.getListenerPort();
+        String path = this.uri;
+        if (location == null || location.isEmpty()) {
+            if (!action.host.isEmpty()) {
+                host = action.host;
+            }           
+            if (action.port > 0) {
+                port = action.port;
+            } else if ("https".equals(action.redirectProto)) {
```
eolivelli: "https" -> constant ?

carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
```diff
@@ -91,12 +93,22 @@ public ClientConnectionHandler(
...
+        this.listenerHost = listenerHost;
+        this.listenerPort = listenerPort;
     }
 
     public SocketAddress getClientAddress() {
         return clientAddress;
     }
 
+    public String getListenerHost() {
+        return listenerHost;
```
eolivelli: maybe this is not needed.

the listener host/port is an internal information useful only inside the local network, and usually the listener will be on 0.0.0.0 so this information cannot be used for constructing an URI

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -386,6 +392,42 @@ private void serveStaticMessage() {
...
+    private void serveRedirect() {
+        DefaultFullHttpResponse res = new DefaultFullHttpResponse(
+                HttpVersion.HTTP_1_1,
+                HttpResponseStatus.valueOf(action.errorcode < 0 ? 302 : action.errorcode) // redirect: 3XX
+        );
+        addCustomResponseHeaders(res);
+        res.headers().set(HttpHeaderNames.CACHE_CONTROL, "no-cache");
+
+        String location = action.redirectLocation;
+        String host = this.connectionToClient.getListenerHost();
```
eolivelli: do not use this info

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -386,6 +392,42 @@ private void serveStaticMessage() {
...
+        int port = this.connectionToClient.getListenerPort();
```
eolivelli: do not use this info (NAT can be in place)

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
eolivelli: I would start with the original Host used by the client, 
we can use this information in case the "Host" header is not available

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
eolivelli: we can use it in case the "Host" header is not available

carapace-server/src/main/java/org/carapaceproxy/server/config/ActionConfiguration.java
```diff
@@ -33,24 +33,29 @@
...
 
     private final String id;
     private final String type;
-    private final String file;
     private final String director;
-    private final int errorcode;    
-    private final List<CustomHeader> customHeaders; // it's a list to keep ordering
+    private final String file;
+    private final int errorcode;
+    private List<CustomHeader> customHeaders = Collections.unmodifiableList(new ArrayList()); // it's a list to keep ordering
```
eolivelli: Collections.unmodifiableList(new ArrayList());   -> Collections.emptyList()

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -160,8 +150,24 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
...
+                ActionConfiguration _action = new ActionConfiguration(id, action, director, file, code)
+                        .setCustomHeaders(customHeaders);
+                String redirectLocation = properties.getProperty(prefix + "redirect.location", "");
+                _action.setRedirectLocation(redirectLocation);
+                if (redirectLocation.isEmpty()) {
+                    _action.setRedirectProto(properties.getProperty(prefix + "redirect.proto", ""));
+                    _action.setRedirectHost(properties.getProperty(prefix + "redirect.host", ""));
+                    _action.setRedirectPort(Integer.parseInt(properties.getProperty(prefix + "redirect.port", "-1")));
+                    _action.setRedirectPath(properties.getProperty(prefix + "redirect.path", ""));
+                }
```
eolivelli: we have to fail in case that we don't have set any of the five parameters

carapace-server/src/test/java/org/carapaceproxy/server/mapper/BasicStandardEndpointMapperTest.java
```diff
@@ -439,4 +427,101 @@ public void testCustomAndDebbugingHeaders() throws Exception {
...
+            // redirect to same domain/uri but with https
+            configuration.put("route.1.id", "r1");
+            configuration.put("route.1.enabled", "true");
+            configuration.put("route.1.match", "regexp .*index\\.html");
+            configuration.put("route.1.action", "a1");
+            configuration.put("action.1.id", "a1");
+            configuration.put("action.1.enabled", "true");
+            configuration.put("action.1.type", "redirect");
+            configuration.put("action.1.code", "301");
+            configuration.put("action.1.redirect.proto", "https");
```
eolivelli: I am expecting here that the client will receive a Location header like:
https://HOST-HEADER/same-original-path

where HOST-HEADER is exactly the value of the "Host" HTTP Header inside the Request

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -386,6 +395,44 @@ private void serveStaticMessage() {
...
+        host = host.split(":")[0];
+        String path = this.uri;
+        if (location == null || location.isEmpty()) {
+            if (!action.host.isEmpty()) {
+                host = action.host;
+            }
+            if (action.port > 0) {
+                port = ":" + action.port;
+            } else if (PROTO_HTTPS.equals(action.redirectProto)) {
+                port = ":" + PROTO_HTTP_DEFAULT_PORT;
```
eolivelli: Do not write default ports like :443 or :80

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -160,8 +150,30 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
...
+                ActionConfiguration _action = new ActionConfiguration(id, action, director, file, code)
+                        .setCustomHeaders(customHeaders);
+                String redirectLocation = properties.getProperty(prefix + "redirect.location", "");
+                _action.setRedirectLocation(redirectLocation);
+                if (redirectLocation.isEmpty()) {
+                    _action.setRedirectProto(properties.getProperty(prefix + "redirect.proto", ""));
+                    _action.setRedirectHost(properties.getProperty(prefix + "redirect.host", ""));
+                    _action.setRedirectPort(Integer.parseInt(properties.getProperty(prefix + "redirect.port", "-1")));
+                    _action.setRedirectPath(properties.getProperty(prefix + "redirect.path", ""));
+                    if (action.equals("redirect") && _action.getRedirectProto().isEmpty() && _action.getRedirectHost().isEmpty()
```
eolivelli: Constants?

PR #67, opened 2019-05-06
Migrating from NPM to Yarn and upgranding dependencies

mino181295:
In this commit I have:

- Migrated NPM to Yarn
- Upgraded the pom of the UI to the newest version of mave-frontend-plugin and added yarnVersion instead of npmVersion to install the dependencies
- Upgraded the UI package.json to have the latest version of each dependency
- Removed BootstrapVue because we are using bootstrap.scss directly from Bootstrap 4.1.3


eolivelli:
well done @mino181295 
merging as soon as Travis is happy

Issue #60, opened 2019-04-29
Handle dynamic certificates in cluster mode


eolivelli:
Certificates are stored on the replicated database (HerdDB) but in case of renew the peer that updates the certificate must notify other peers, this way every peer will serve the right certificate.

PR #70, opened 2019-05-08
Handle dynamic certificates in cluster mode #60

Pols95:


carapace-server/src/main/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandler.java
```diff
@@ -187,6 +188,17 @@ public void stop() {
         callbacksExecutor.shutdown();
     }
 
+    @Override
+    public void executeInMutex(String mutexId, Runnable runnable) throws Exception {        
+        InterProcessMutex mutex = new InterProcessMutex(client, "/proxy/mutex/" + mutexId);
+        try {
+            mutex.acquire();
```
eolivelli: in case we cannot acquire the Mutex we can simply exit without reporting an error.
This should be stated in higher level interface GroupMembershipHandler.java

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
```diff
@@ -310,19 +329,26 @@ private KeyPair loadKeyPair(String pk) throws Exception {
...
+    private boolean saveKeyPair(KeyPair pair, String pk, boolean update) {
         try (Connection con = datasource.getConnection(); PreparedStatement psInsert = con.prepareStatement(INSERT_INTO_KEYPAIR_TABLE); PreparedStatement psUpdate = con.prepareStatement(UPDATE_KEYPAIR_TABLE)) {
             String privateKey = base64EncodeKey(pair.getPrivate());
             String publicKey = base64EncodeKey(pair.getPublic());
-            psUpdate.setString(1, privateKey);
-            psUpdate.setString(2, publicKey);
-            psUpdate.setString(3, pk);
-            if (psUpdate.executeUpdate() == 0) {
+            boolean updateDone = false;
+            if (update) {
```
eolivelli: what does 'update = false' mean ?
this method won't do anything

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
```diff
@@ -376,4 +416,48 @@ public void saveCertificate(CertificateData cert) {
...
+    }
+
+    @Override
+    public String loadAcmeChallengeToken(String id) {
+        try (Connection con = datasource.getConnection()) {
+            try (PreparedStatement ps = con.prepareStatement(SELECT_FROM_ACME_CHALLENGE_TOKENS_TABLE)) {
+                ps.setString(1, id);
+                try (ResultSet rs = ps.executeQuery()) {
+                    if (rs.next()) {
+                        return rs.getString(2);
```
eolivelli: why returning the 2nd element ? you can narrow the query an return only that column

carapace-server/src/main/java/org/carapaceproxy/server/certiticates/DynamicCertificatesManager.java
```diff
@@ -80,11 +81,17 @@
     private ConfigurationStore store;
     private volatile int period = 0; // in seconds
     private volatile int keyPairsSize = DEFAULT_KEYPAIRS_SIZE;
+    private GroupMembershipHandler groupMembershipHandler;
 
     public void setConfigurationStore(ConfigurationStore configStore) {
         this.store = configStore;
     }
 
+    public void setGroupMembershipHandler(GroupMembershipHandler groupMembershipHandler) {
```
eolivelli: this is not a simply "setter"
it is better to rename it to something like 'attachGroupMembershipHandler'

carapace-server/src/main/java/org/carapaceproxy/server/certiticates/DynamicCertificatesManager.java
```diff
@@ -310,23 +333,52 @@ public void setStateOfCertificate(String id, DynamicCertificateState state) {
...
+            } catch (Exception err) {
+                LOG.log(Level.SEVERE, "Cannot apply new configuration");
+            }
         }
-        return Base64.getDecoder().decode(data.getChain());
     }
 
+    private void reloadCertificatesFromDB() {
+        try {
+            ConcurrentHashMap _certificates = new ConcurrentHashMap();
```
eolivelli: use generics

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
Pols95: Maybe it's safer to allow only KeyPairs inserts?

eolivelli: Yes, probably

Pols95:
All unit tests should pass now.. Waiting for tomorrow for real test of the cluster.

carapace-server/src/main/java/org/carapaceproxy/server/certiticates/DynamicCertificateState.java
```diff
@@ -0,0 +1,35 @@
...
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+
+ */
+package org.carapaceproxy.server.certiticates;
+
+/**
+ *
```
Pols95: javadoc

Issue #78, opened 2019-05-20
UI: Backends view shows backend_id:0 instead of host:port


Pols95:
In some cases, there are "duplicated" rows for backends with backend_id:0 instead of host:port.. should be skipped.

PR #84, opened 2019-05-22
EndpointMapper: Backends availability cheked according to backendid instead of host:port #78

Pols95:
In StandardEndpoinMapper backendHealthManager.isAvailable() is called with backendId instead of hostname:port as expected by BackendHealthManager.

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
@@ -414,7 +414,7 @@ public class StandardEndpointMapper extends EndpointMapper {
                     }
 
                     BackendConfiguration backend = this.backends.get(backendId);
-                    if (backend != null && backendHealthManager.isAvailable(backendId)) {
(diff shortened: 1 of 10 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -414,7 +414,7 @@ public MapResult map(HttpRequest request, String userId, String sessionId, Backe
                     }
 
                     BackendConfiguration backend = this.backends.get(backendId);
-                    if (backend != null && backendHealthManager.isAvailable(backendId)) {
+                    if (backend != null && backendHealthManager.isAvailable(backend.getHostPort())) {
```
eolivelli: this fix seems unrelated to the title

Pols95: Whenever a backend availability is checked a new BackendHealthStatus object is created. This is because the id is always new due to backends list resetting performed in BackendHealthManager.run according to the configuration (which consider host:port as id). The conseguences is that the new BackendHealthStatus with default reportedAsUnreachable = false  seems to be always consider as available.

PR #87, opened 2019-05-22
Enable ZK Configuration and ACL

eolivelli:


eolivelli:
Tests are missing, this is not ready

eolivelli:
@dmercuriali  @Pols95 I have added a minimal test about ACL.

The test starts ZooKeeper server with SASLAuthenticationProvider and sets a global JAAS configuration file.
The Client and the Server just by seeing that file (using the system property java.security.auth.login.config) start using JAAS for auth.

If the auth is not correct the test fails because you cannot set the requested ACL if you are not authenticated.
In a real setup we will be using Kerberos and not MD5 (that is not secure, it is "fuffa") but this is up to ZooKeeper, we do not need any special handling of Kerberos, it is enabled just by setting the correct parameters on JAAS configuration files.

carapace-server/src/test/java/org/carapaceproxy/cluster/impl/ZookKeeperACLTest.java
```diff
@@ -0,0 +1,177 @@
...
+                    ZooKeeperGroupMembershipHandler peer2 = new ZooKeeperGroupMembershipHandler(testingServer.getConnectString(),
+                            6000, true /*acl */, peerId2, Collections.EMPTY_MAP, new Properties())) {
+                peer1.start();
+                peer2.start();
+                List<String> peersFrom1 = peer1.getPeers();
+                List<String> peersFrom2 = peer2.getPeers();
+                assertEquals(Arrays.asList(peerId1, peerId2), peersFrom1);
+                assertEquals(Arrays.asList(peerId1, peerId2), peersFrom2);
+
+                AtomicInteger eventFired2 = new AtomicInteger();;
```
dmercuriali: Double ;;

carapace-server/src/test/java/org/carapaceproxy/cluster/impl/ZookKeeperACLTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+                    @Override
+                    public void reconnected() {
+
+                    }
+                });
+
+                peer1.fireEvent("foo");
+
+                for (int i = 0; i < 10; i++) {
```
dmercuriali: Maybe use TestUtils#waitForCondition. It's basically the same thing but we avoi some clutter

carapace-server/src/test/java/org/carapaceproxy/cluster/impl/ZookKeeperACLTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+                        @Override
+                        public void reconnected() {
+
+                        }
+                    });
+
+                    peer1.fireEvent("foo");
+
+                    for (int i = 0; i < 10; i++) {
```
dmercuriali: same as above

carapace-server/src/test/java/org/carapaceproxy/cluster/impl/ZookKeeperACLTest.java
eolivelli: done

carapace-server/src/test/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandlerTest.java
```diff
@@ -107,17 +109,14 @@ public void reconnected() {
...
-                for (int i = 0; i < 10; i++) {
-                    if (eventFired2.get() >= 1) {
-                        break;
-                    }
-                    Thread.sleep(100);
-                }
+                TestUtils.waitForCondition(() -> {
+                    return eventFired2.get() >= 1;
+                }, 100);
                 assertTrue(eventFired2.get() >= 1);
```
dmercuriali: waitForCondition will fail if the condition is not met in time so this assert is redundant

carapace-server/src/test/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandlerTest.java
```diff
@@ -150,13 +146,10 @@ public void reconnected() {
...
-                    for (int i = 0; i < 10; i++) {
-                        if (eventFired2.get() >= 1
-                                && eventFired3.get() >= 1) {
-                            break;
-                        }
-                        Thread.sleep(100);
-                    }
+                    TestUtils.waitForCondition(() -> {
+                        return (eventFired2.get() >= 1
+                                && eventFired3.get() >= 1);
```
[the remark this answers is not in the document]
eolivelli: done

eolivelli:
@dmercuriali  I have fixed the test.
thank you

PR #107, opened 2019-06-05
Add systemd script

hamadodene:
If you need to configure the service to run at startup, you can use this systemd script.

- edit carapace.service and insert correct "User", "Group" and "installation path"
- copy carapace.service in /etc/systemd/system
- systemctl enable carapace.service 
- systemctl start carapace.service (To start carapace)
- systemctl stop carapace.service (To stop carapace)
- systemctl restart carapace.service (To restart carapace)


eolivelli:
great.
let's wait for CI

PR #114, opened 2019-06-17
Auth not working in database mode

Pols95:
Now userrealm class is loaded at boot as "static configuration"


carapace-server/src/main/java/org/carapaceproxy/server/HttpProxyServer.java
```diff
@@ -534,11 +541,7 @@ public Listeners getListeners() {
 
     public RuntimeServerConfiguration buildValidConfiguration(ConfigurationStore simpleStore) throws ConfigurationNotValidException {
         RuntimeServerConfiguration newConfiguration = new RuntimeServerConfiguration();
-        newConfiguration.configure(simpleStore);
-        // creating a mapper validates the configuration
-        buildMapper(newConfiguration.getMapperClassname(), simpleStore);
```
eolivelli: why are you dropping these lines ?

Issue #122, opened 2019-06-20
BUG: BackendHealthStatus thoubles with probes


Pols95:
In test-cluster, probes always returns httpsResponse==null

Pols95:
By default healthmanager execution manager is 0...

PR #123, opened 2019-06-20
BUG: BackendHealthStatus thoubles with probes #122

Pols95:


carapace-server/src/main/java/org/carapaceproxy/server/RuntimeServerConfiguration.java
```diff
@@ -272,7 +273,7 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
             tryConfigureFilter(i, properties);
         }
 
-        healthProbePeriod = getInt("healthmanager.period", 0, properties);
+        healthProbePeriod = getInt("healthmanager.period", DEFAULT_PERIOD, properties);
```
eolivelli: You will break tests in an unpredictable way.
The health manager must stay off in tests

PR #137, opened 2019-07-11
NPE: HTTPS Admin Interface have default certificate password NULL

hamadodene:


carapace-server/src/main/java/org/carapaceproxy/server/HttpProxyServer.java
```diff
@@ -130,7 +130,7 @@
     private String adminAdvertisedServerHost = adminServerHost; // hostname to access API/UI
     private int adminServerHttpsPort = -1;
     private String adminServerCertFile;
-    private String adminServerCertFilePwd;
+    private String adminServerCertFilePwd="";
```
eolivelli: Does this really fix your problem ?

I guess you should take a look at where we are reading the value for adminServerCertFilePwd

hamadodene: We are reading the value in applyStaticConfiguration: 
adminServerCertFilePwd = properties.getProperty("https.admin.sslcertfilepassword", adminServerCertFilePwd);

But if the propertie is not set in configuration file the value would be null

eolivelli: ok

Issue #140, opened 2019-10-01
Dynamic Certificates reloading from DB discards certificate "mode"


hamadodene:
I uploaded my certificate in carapace with the manual mode, but Carapace attempts to renew it

```
Oct 01, 2019 2:58:26 PM org.carapaceproxy.server.certiticates.DynamicCertificatesManager certificatesLifecycle
INFO: Certificate ISSUING process for domain: MYDOMAIN STARTED.
Oct 01, 2019 2:58:26 PM org.carapaceproxy.server.certiticates.DynamicCertificatesManager certificatesLifecycle
SEVERE: Error while handling dynamic certificate for domain MYDOMAIN

SEVERE: Error while handling dynamic certificate for domain MYDOMAIN
org.shredzone.acme4j.exception.AcmeServerException: Error creating new order :: Cannot issue for "MYDOMAIN": Name does not end in a public suffix
        at org.shredzone.acme4j.connector.DefaultConnection.throwAcmeException(DefaultConnection.java:517)
        at org.shredzone.acme4j.connector.DefaultConnection.performRequest(DefaultConnection.java:442)
        at org.shredzone.acme4j.connector.DefaultConnection.sendSignedRequest(DefaultConnection.java:350)
        at org.shredzone.acme4j.connector.DefaultConnection.sendSignedRequest(DefaultConnection.java:157)
        at org.shredzone.acme4j.OrderBuilder.create(OrderBuilder.java:272)
        at org.carapaceproxy.server.certiticates.ACMEClient.createOrderForDomain(ACMEClient.java:102)
        at org.carapaceproxy.server.certiticates.DynamicCertificatesManager.certificatesLifecycle(DynamicCertificatesManager.java:229)
        at org.carapaceproxy.server.certiticates.DynamicCertificatesManager.lambda$run$0(DynamicCertificatesManager.java:210)
        at org.carapaceproxy.cluster.impl.ZooKeeperGroupMembershipHandler.executeInMutex(ZooKeeperGroupMembershipHandler.java:297)
        at org.carapaceproxy.server.certiticates.DynamicCertificatesManager.run(DynamicCertificatesManager.java:208)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
[...]

PR #141, opened 2019-10-01
Dynamic Certificates reloading from DB discards certificate mode #140

Pols95:


carapace-server/src/main/java/org/carapaceproxy/server/certiticates/DynamicCertificatesManager.java
```diff
@@ -452,7 +455,9 @@ private void reloadCertificatesFromDB() {
         try {
             Map<String, CertificateData> _certificates = new ConcurrentHashMap<>();
             for (String domain : certificates.keySet()) {
```
eolivelli: you should loop over the  entrySet

PR #163, opened 2019-12-06
Default route selected with all backends down

pv3nturi:


carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -456,13 +454,14 @@ public MapResult map(HttpRequest request, String userId, String sessionId, Backe
                                 .setCustomHeaders(customHeaders);
                     }
                 }
+                // none of selected backends available
+                if (!selectedBackends.isEmpty()) {
+                    return MapResult.INTERNAL_ERROR(route.getId());
```
pv3nturi: Not sure which is better: NOT_FOUND or INTERNAL_ERROR
@eolivelli

aluccaroni: @pv3nturi it should return "http 503 (service not available)". If that't not possible, lets return http 500 (INTERNAL_ERROR, right?) and fix that in a future issue

pv3nturi: "http 503 (service not available)" is currently unavailable... let's fix it in the future

PR #167, opened 2019-12-11
Better EndpointConnection lifecycle handling:

eolivelli:
- ensure that changes to the state of EndpointConnectionImpl happen inside the Endpoint event loop
- track status of writes to the client in RequestHandler (useful for heapdumps)
- release connection to the endpoint as soon as we receive the LastHttpContent from the backend, this saves resource in case of slow or stuck clients (non writable socket to the client)

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/client/EndpointConnection.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/client/EndpointConnection.java
@@ -26,7 +26,7 @@ import org.carapaceproxy.server.RequestHandler;
 
 /**
  * A Connection to a specific endpoint. Connections are pooled and so they have
- * to be returned to the pool
+ * to be returned to the pool. A connection can be bound to at most one RequestHandler at a time.
  *
  * @author enrico.olivelli
  */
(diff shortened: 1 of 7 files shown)


carapace-server/src/main/java/org/carapaceproxy/client/EndpointConnection.java
```diff
@@ -26,20 +26,38 @@
...
+ * to be returned to the pool. A connection can be bound to at most one RequestHandler at a time.
  *
  * @author enrico.olivelli
  */
 public interface EndpointConnection {
 
     public EndpointKey getKey();
 
+    /**
+     * Start a requrest.
```
pv3nturi: request

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -257,6 +263,7 @@ public void start() {
     }
 
     void continueClientRequest(HttpContent httpContent) {
+        LOG.log(Level.INFO, "continueClientRequest {0}", this);
```
eolivelli: debug to remove

PR #171, opened 2019-12-13
Support for OCSP stapling, failing test case

eolivelli:
- Upgrade Netty tcboring to 2.0.28Final, with 2.0.27Final tests are crashing on my Mac
- Add a failing test case that certifies that CSP stapling does not work


eolivelli:
Creating a test is not possible with selft-signed certificates.
I am closing now this PR

PR #178, opened 2019-12-18
Make tests pass on JDK12+ use Unsafe

eolivelli:
use Unsafe to set final field values


eolivelli:
@pv3nturi I have added another commit, because I had a failure on MacOS, on MacOs  you cannot use address 0.0.0.1

PR #187, opened 2020-01-18
UI: Navigation bar integration

pv3nturi:
- Added a top Navbar (in order to split/support future features like users-management, logout, preferences and so on...)
- Changed the App font to Exo 2 (https://fonts.google.com/specimen/Exo+2)


pv3nturi:
carapace-ui

carapace-ui/src/main/webapp/public/index.html
```diff
@@ -8,6 +8,7 @@
     <title>Carapace</title>
     <link rel="shortcut icon" href="<%= BASE_URL %>assets/logo.png" type="image/x-icon">
     <link rel="icon" href="<%= BASE_URL %>assets/logo.png" type="image/x-icon">
+    <link href='https://fonts.googleapis.com/css?family=Exo 2' rel='stylesheet'>
```
mino181295: I think we should keep this in the resources in assets folder to avoid an external call

mino181295:
Just an imprvement, not a big deal.

+1 to me

Issue #185, opened 2020-01-16
Route error messages customization


pv3ntur1:
We need a way to customize error responses (error messages) from routes.
The solution is to bind to the route an "error-action" of type "static" (linked to corresponding error page.) to perform in case of error (the code - ex 500- is defined in the action.code)

Example config:
```
route.x.action=action
route.x.enabled=true
route.x.id=route_x
route.x.match=somethin
route.x.erroraction=action_error

action.x.id=action_error
action.x.type=static
action.x.resource=xxxxx
action.x.code=500
```

NOTE: the error "404 - Not found" is already customizable globally.

hamadodene:
Note, if action linked to inexistent file,  we need to generate "not found error" (404).
I perfomed a test and I abtain this error:

```
SEVERE: Cannot load resource /data/carapace/default-error-pages/inexistent_page.html: java.io.IOException: cannot load resource /data/carapace/default-error-pages/inexistent_page.html, path must start with file: or classpath:
Jan 27, 2020 12:49:05 PM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException
WARNING: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.NullPointerException
        at org.carapaceproxy.server.StaticContentsManager.buildResponse(StaticContentsManager.java:59)
        at org.carapaceproxy.server.RequestHandler.serveStaticMessage(RequestHandler.java:402)
        at org.carapaceproxy.server.RequestHandler.clientRequestFinished(RequestHandler.java:327)
        at org.carapaceproxy.server.ClientConnectionHandler.channelRead0(ClientConnectionHandler.java:160)
        at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:374)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:360)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:352)
        at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:326)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:300)
[...]

PR #189, opened 2020-01-29
Route error messages customization #185

pv3nturi:
- Now it's possible to attach a custom action/message to routes, in order to serve the proper response/code in case of error (see #185).

- Fixed a NPE in response building by static content (now will be return 404 for resources unable to find)

- Added Error-Action column to Routes UI

- Fixed custom headers support for Error Actions

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/EndpointMapper.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/EndpointMapper.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import org.carapaceproxy.client.EndpointKey;
 import org.carapaceproxy.configstore.ConfigurationStore;
 import org.carapaceproxy.server.RequestHandler;
+import org.carapaceproxy.server.StaticContentsManager;
 import org.carapaceproxy.server.backends.BackendHealthManager;
 import org.carapaceproxy.server.certificates.DynamicCertificatesManager;
 import org.carapaceproxy.server.config.ActionConfiguration;
@@ -53,12 +54,12 @@ public abstract class EndpointMapper {
 
     public abstract MapResult map(HttpRequest request, String userId, String sessionId, BackendHealthManager backendHealthManager, RequestHandler requestHandler);
 
-    public MapResult mapInternalError(HttpRequest request, String routeid) {
-        return MapResult.INTERNAL_ERROR(routeid);
+    public SimpleHTTPResponse mapPageNotFound(String routeid) {
+        return SimpleHTTPResponse.NOT_FOUND(StaticContentsManager.DEFAULT_NOT_FOUND);
     }
 
-    public MapResult mapPageNotFound(HttpRequest request, String routeid) {
-        return MapResult.NOT_FOUND(routeid);
+    public SimpleHTTPResponse mapInternalError(String routeid) {
+        return SimpleHTTPResponse.INTERNAL_ERROR(StaticContentsManager.DEFAULT_INTERNAL_SERVER_ERROR);
     }
 
     public void endpointFailed(EndpointKey key, Throwable error) {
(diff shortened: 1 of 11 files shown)


carapace-server/src/main/java/org/carapaceproxy/EndpointMapper.java
```diff
@@ -53,11 +53,11 @@
 
     public abstract MapResult map(HttpRequest request, String userId, String sessionId, BackendHealthManager backendHealthManager, RequestHandler requestHandler);
 
-    public MapResult mapDefaultInternalError(HttpRequest request, String routeid) {
+    public MapResult mapInternalError(HttpRequest request, String routeid) {
```
eolivelli: Maybe it is better to create a new class SimpleHTTPResponse that only contains the HTTP status code,  message, headers.....
This way it is clearer that the outcome cannot be *any* kind of action but only a static error code

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -490,29 +491,23 @@ public MapResult mapInternalError(HttpRequest request, String routeid) {
...
-    public MapResult mapPageNotFound(HttpRequest request, String routeid) {
+    public SimpleHTTPResponse mapPageNotFound(String routeid) {
         // custom global
         if (defaultNotFoundAction != null) {
             ActionConfiguration actionError = actions.get(defaultNotFoundAction);
             if (actionError != null) {
-                return new MapResult(null, 0, Action.NOTFOUND, routeid)
-                        .setResource(actionError.getFile())
-                        .setErrorcode(actionError.getErrorcode())
-                        .setCustomHeaders(actionError.getCustomHeaders());
```
eolivelli: it looks like we are losing actionError.getCustomHeaders()

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -490,29 +491,23 @@ public MapResult mapInternalError(HttpRequest request, String routeid) {
             errorAction = actions.get(defaultInternalErrorAction);
         }
         if (errorAction != null) {
-            return new MapResult(null, 0, Action.INTERNAL_ERROR, routeid)
-                    .setResource(errorAction.getFile())
-                    .setErrorcode(errorAction.getErrorcode())
-                    .setCustomHeaders(errorAction.getCustomHeaders());
```
eolivelli: it looks like we are losing actionError.getCustomHeaders()

carapace-server/src/main/java/org/carapaceproxy/SimpleHTTPResponse.java
```diff
@@ -0,0 +1,69 @@
...
+public class SimpleHTTPResponse {
+
+    private final int errorcode;
+    private final String resource;
+    private final List<CustomHeader> customHeaders;
+
+    public SimpleHTTPResponse(int errorcode, String resource, List<CustomHeader> customHeaders) {
+        this.errorcode = errorcode;
+        this.resource = resource;
+        this.customHeaders = Collections.unmodifiableList(customHeaders == null ? new ArrayList() : customHeaders);
```
eolivelli: nit:
`Collections.unmodifiableList(customHeaders == null ? new ArrayList()`

why not this form ? no memory allocations
`Collections.emptyList()`

carapace-server/src/main/java/org/carapaceproxy/server/config/ActionConfiguration.java
```diff
@@ -40,7 +40,7 @@
     private final String director;
     private final String file;
     private final int errorcode;
-    private List<CustomHeader> customHeaders = Collections.emptyList(); // it's a list to keep ordering
+    private List<CustomHeader> customHeaders = Collections.EMPTY_LIST; // it's a list to keep ordering
```
eolivelli: why using the unchecked EMPTY_LIST ? isn't Collections.<>emptyList() better ?

PR #204, opened 2020-03-26
Configuration entries max size fixed

pv3nturi:
In order to manage a greater amount of certificates we need to drop the limit of 100 entries for property index in configuration file.

By this improve, now configuration properties are fetched by iterating from index 0 to the max for that property prefix-name.

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/RuntimeServerConfiguration.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/RuntimeServerConfiguration.java
@@ -289,10 +289,9 @@ public class RuntimeServerConfiguration {
     }
 
     private void tryConfigureCertificates(ConfigurationStore properties) throws ConfigurationNotValidException {
-        String prefix = "certificate.";
-        int max = properties.nextIndexFor(prefix);
-        for (int i = 0; i < max; i++) {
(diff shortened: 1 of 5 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/RuntimeServerConfiguration.java
```diff
@@ -294,72 +288,81 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
...
-                LOG.log(Level.INFO,
-                        "Configuring SSL certificate {0}: hostname={1}, file={2}, password={3}, mode={4}",
-                        new Object[]{prefix, hostname, file, pw, mode}
-                );
-                SSLCertificateConfiguration config = new SSLCertificateConfiguration(hostname, file, pw, _mode);
-                if (config.isAcme()) {
-                    config.setDaysBeforeRenewal(daysBeforeRenewal);
+    private void tryConfigureCertificates(ConfigurationStore properties) throws ConfigurationNotValidException {
+        String prefix = "certificate.";
+        int max = properties.nextIndexFor(prefix);
```
eolivelli: why "nextIndexFor" ?

what about having some "getMaxIndexForPrefix".... ?
with specific unit tests

PR #206, opened 2020-04-19
HTTP/2.0 Integration

pv3nturi:


carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -852,4 +858,11 @@ public boolean isSecure() {
         return connectionToClient.isSecure();
     }
 
+    private static void setStreamIdFromRequest(HttpObject response, HttpRequest request) {
+        if (response instanceof HttpResponse) {
+            String streamId = request.headers().get(HttpConversionUtil.ExtensionHeaderNames.STREAM_ID.text());
```
eolivelli, line 863: is this visible to the end user ?

pv3nturi:
We gonna support HTTP/2.0 by using Reactor Netty

PR #230, opened 2020-11-27
Netty eventloop deadlock on connections release

pv3nturi:
Now connections releasing and recreation for the pool is done throw a dedicated thread pool instead of netty eventloop.

Sistem-property for threadpool size: _**carapace.connectionsmanager.returnconnectionthreadpool.size**_ (default **10**; 0 for rollback)

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
@@ -87,11 +87,13 @@ public class ConnectionsManagerImpl implements ConnectionsManager, AutoCloseable
             "stuck_requests_total",
             "stuck requests, this requests will be killed").register();
 
-    private static final int CONNECTIONS_MANAGER_THREADS_POOL_SIZE = Integer.getInteger("carapace.connectionsmanager.threadspool.size", 10);
-    private final ExecutorService threadsPool;
+    private static final int CONNECTIONS_MANAGER_RETURN_CONNECTION_THREAD_POOL_SIZE = Integer.getInteger("carapace.connectionsmanager.returnconnectionthreadpool.size", 10);
+    private final ExecutorService returnConnectionThreadPool;
 
     public void returnConnection(EndpointConnectionImpl con) {
-        threadsPool.submit(() -> {
+        // We need to perform returnObject in dedicated thread in order to avoid deadlock in Netty evenLoop
+        // in case of connection re-creation
+        returnConnectionThreadPool.submit(() -> {
             LOG.log(Level.FINE, "returnConnection:{0}", con);
             connections.returnObject(con.getKey(), con);
         });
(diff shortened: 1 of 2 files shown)


carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
```diff
@@ -86,9 +87,14 @@
             "stuck_requests_total",
             "stuck requests, this requests will be killed").register();
 
+    private static final int CONNECTIONS_MANAGER_THREADS_POOL_SIZE = Integer.getInteger("carapace.connectionsmanager.threadspool.size", 10);
```
eolivelli, line 90: this is not good, as we know that this is used by commons pool to handle new connections

it is okay for an hotfix, but I expect this size to be computed from the number of maximum open connections

carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
```diff
@@ -86,9 +87,14 @@
...
+    private final ExecutorService threadsPool;
+
     public void returnConnection(EndpointConnectionImpl con) {
-        LOG.log(Level.FINE, "returnConnection:{0}", con);
-        connections.returnObject(con.getKey(), con);
+        threadsPool.submit(() -> {
```
eolivelli, line 94: please add a comment

carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
```diff
@@ -219,6 +225,7 @@ public final void applyNewConfiguration(RuntimeServerConfiguration configuration
 
     public ConnectionsManagerImpl(RuntimeServerConfiguration configuration, BackendHealthManager backendHealthManager) {
         this.scheduler = Executors.newSingleThreadScheduledExecutor();
+        this.threadsPool = Executors.newFixedThreadPool(CONNECTIONS_MANAGER_THREADS_POOL_SIZE);
```
eolivelli, line 228: what about "returnConnectionThreadPool"?

eolivelli (changes requested):

carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
```diff
@@ -219,6 +227,8 @@ public final void applyNewConfiguration(RuntimeServerConfiguration configuration
 
     public ConnectionsManagerImpl(RuntimeServerConfiguration configuration, BackendHealthManager backendHealthManager) {
         this.scheduler = Executors.newSingleThreadScheduledExecutor();
+        LOG.log(Level.INFO, "Reading carapace.connectionsmanager.returnconnectionthreadpool.size=" + CONNECTIONS_MANAGER_RETURN_CONNECTION_THREAD_POOL_SIZE);
+        this.returnConnectionThreadPool = Executors.newFixedThreadPool(CONNECTIONS_MANAGER_RETURN_CONNECTION_THREAD_POOL_SIZE);
```
eolivelli, line 231: this thread pool is not shutdown

eolivelli (changes requested):

carapace-server/src/main/java/org/carapaceproxy/client/impl/ConnectionsManagerImpl.java
```diff
@@ -302,6 +312,13 @@ public void close() {
         connections.close();
         group.shutdownGracefully();
         eventLoopForOutboundConnections.shutdownGracefully();
+
+        try {
+            returnConnectionThreadPool.shutdown();
+            returnConnectionThreadPool.awaitTermination(10, TimeUnit.SECONDS);
+        } catch (InterruptedException ex) {
+            LOG.severe("Error wating for returnConnectionThreadPool termination: " + ex);
```
eolivelli, line 320: you are missing Thread.currentThread.interrupt()

eolivelli (changes requested):

eolivelli (approved):
LGTM

eolivelli:
the approach may work.
there is a risk that an unhealthy backend uses all of the threads in this new thread pool and connections are no more released.

we should allow to disable this feature, like setting the configurable value to 0 (in this case use MoreExecutors.directExecutor()

PR #232, opened 2020-12-06
Filter requests based on server ip

hamadodene:
In Route configuration,  it might be useful to separate the traffic based on the ip used to forward the request.
This is because there can be dedicated ip's  for a particular customer.

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
@@ -157,7 +157,7 @@ public class RequestHandler implements MatchingContext {
     InetSocketAddress getLocalAddress() {
         return (InetSocketAddress) channelToClient.channel().localAddress();
     }
-
+    
     private void fireRequestFinished() {
         Runnable handler = onRequestFinished.getAndSet(null);
         if (handler != null) {
@@ -570,7 +570,7 @@ public class RequestHandler implements MatchingContext {
         connectionToClient.lastHttpContentSent(this);
         requestsLogger.logRequest(this);
     }
-
+    
     @Override
     public int hashCode() {
         int hash = 7;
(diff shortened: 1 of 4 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -836,6 +837,9 @@ public String getProperty(String name) {
                     return request.method().name();
                 case PROPERTY_CONTENT_TYPE:
                     return request.headers().get(javax.ws.rs.core.HttpHeaders.CONTENT_TYPE, "");
+                case PROPERTY_SERVER_IP:
```
pv3nturi, line 840: maybe this should be PROPERTY_LISTENER_ADDRESS and the property below PROPERTY_LISTENER_HOST_PORT

pv3nturi (changes requested):

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -821,7 +821,8 @@ public void badErrorOnRemote(Throwable cause) {
     public static final String PROPERTY_CONTENT_TYPE = "request.content-type";
     public static final String PROPERTY_HEADERS = "request.headers.";
     private static final int HEADERS_SUBSTRING_INDEX = PROPERTY_HEADERS.length();
-    public static final String PROPERTY_LISTENER_ADDRESS = "listener.address";
+    public static final String PROPERTY_LISTENER_HOST_PORT = "listener.address";
+    public static final String PROPERTY_LISTENER_IPADDRESS = "request.serverip";
```
pv3nturi, line 825: The value of the properties has to be the same of the variable imho:

- PROPERTY_LISTENER_HOST_PORT = "listener.hostport"
- PROPERTY_LISTENER_IPADDRESS = "listener.idaddress"

pv3nturi (changes requested):
Please correct the properties names, 'cause seems the same thing.

pv3nturi:
LGTM!
@hamadodene github actions still fail.

pv3nturi:
I run all tests in local and passed. So a gonna merge.

pv3nturi (approved):

PR #234, opened 2020-12-15
Fix connections release leaks

pv3nturi:
Some fix in order to properly release connections:

- during EndpointConnectionImpl#sendRequest from clients, in case of errorSendingRequest (before starting writing to the server) we need to release the connection immediately or none will do
- when RequestHandler#releaseConnectionToEndpoint is called either by server and client, whether the first one cannot actually release the connection, the second one won't perform the release anymore because of the compareAndSet which set the connection to null even if the release is delayed.
- Added a test case to reproduce the case of a non released connection on errorSendingRequest

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
@@ -19,6 +19,7 @@
  */
 package org.carapaceproxy.client.impl;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
@@ -96,6 +97,8 @@ public class EndpointConnectionImpl implements EndpointConnection {
     private final Gauge.Child openConnectionsStats;
     private final Gauge.Child activeConnectionsStats;
     private final Counter.Child requestsStats;
+
+    private boolean forceErrorOnRequest = false;
 
     private static enum ConnectionState {
         IDLE,
@@ -180,7 +183,7 @@ public class EndpointConnectionImpl implements EndpointConnection {
             throw new IOException(err);
         }
         if (forcedInvalid) {
-            throw new IOException("Cannot connect to " + key+" (already invalidated)");
+            throw new IOException("Cannot connect to " + key + " (already invalidated)");
         }
 
         channelToEndpoint = connectFuture.channel();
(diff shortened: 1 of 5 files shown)


carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
```diff
@@ -200,18 +200,19 @@ public void sendRequest(HttpRequest request, RequestHandler clientSidePeerHandle
...
         }
         checkHandler(null);
+        if (!changeExpectedStateTo(ConnectionState.REQUEST_SENT, ConnectionState.IDLE)) {
+            LOG.log(Level.SEVERE, "bad status ! " + this + ", handler is " + clientSidePeerHandler);
+            throw new IllegalStateException("bad status ! " + this);
+        }
         if (!channelToEndpoint.isOpen() || forcedInvalid) {
             LOG.log(Level.SEVERE, "sendRequest " + request.getClass() + " failed, choosen connection is not valid");
+            changeExpectedStateTo(ConnectionState.RELEASABLE, ConnectionState.REQUEST_SENT);
             clientSidePeerHandler.errorSendingRequest(EndpointConnectionImpl.this, new Exception("no more connected").fillInStackTrace());
```
pv3nturi, line 210: The state has to be RELEASABLE here in order to properly close the connection.

carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
```diff
@@ -235,7 +236,6 @@ public void sendRequest(HttpRequest request, RequestHandler clientSidePeerHandle
                     RequestHandler _clientSidePeerHandler = clientSidePeerHandler;
                     if (!future.isSuccess()) {
                         LOG.log(Level.INFO, this + " sendRequest " + request.getClass() + " failed", future.cause());
-                        changeStateTo(ConnectionState.RELEASABLE, ConnectionState.REQUEST_SENT);
```
pv3nturi, line 238: Whether the server callback runs first the state might be already set to DALAYED_RELEASE. So This is useless.
Whether the server callback runs last it will perform a release on a yet closed channel, throwing an error.

carapace-server/src/main/java/org/carapaceproxy/server/RequestHandler.java
```diff
@@ -669,11 +669,11 @@ private void addCustomResponseHeaders(HttpObject msg, List<CustomHeader> customH
     }
 
     private boolean releaseConnectionToEndpoint(boolean forceClose, EndpointConnection current) {
-        if (connectionToEndpoint.compareAndSet(current, null)) {
+        if (connectionToEndpoint.get().equals(current)) {
```
pv3nturi, line 672: The first one ho performs releaseConnectionToEndpoint set the connection to null even if the release won't be performed.
The second one, who performs delayed release, cannot release 'cause current != null and the connection won't be never closed

carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
```diff
@@ -304,31 +302,36 @@ public void sendLastHttpContent(LastHttpContent msg, RequestHandler clientSidePe
...
-                                LOG.log(Level.SEVERE, "sendLastHttpContent finished without " + ConnectionState.REQUEST_SENT + " state: recovery");
-                                recover = true;
-                            }
-                            clientSidePeerHandler.lastHttpContentSent();
-                            if (recover) {
-                                if (changeStateTo(ConnectionState.RELEASABLE, ConnectionState.DELAYED_RELEASE)) {
-                                    LOG.log(Level.INFO, "recovering DELAYED_RELEASE " + this);
-                                    release(false, clientSidePeerHandler);
-                                }
+                    if (future.isSuccess()) {
```
pv3nturi, line 305: Inverted if

eolivelli (commented):
You aren't adding tests....

nicoloboschi (approved):
Great work, +1

pv3nturi:
All tests pass in local... Failing tests has to be fixed asap in another issue imho

PR #237, opened 2020-12-21
Logs for connection lifecycle debug

pv3nturi:
Added a centralized logger (CarapaceLogger) for debug logging.
Added some debug logs for keep tracking connections lifecycle.


carapace-server/src/main/java/org/carapaceproxy/server/RuntimeServerConfiguration.java
```diff
@@ -298,6 +299,9 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
         ocspStaplingManagerPeriod = properties.getInt("ocspstaplingmanager.period", 0);
         LOG.info("ocspstaplingmanager.period=" + ocspStaplingManagerPeriod);
 
+        boolean loggingLevelFineVisible = properties.getBoolean("logging.level.fine.visible", false);
```
nicoloboschi: visible? it looks weird in this context

maybe we can add a "connection debug mode" where we log the lifecycle of the connection

nicoloboschi (changes requested):

nicoloboschi (approved):
+1

PR #242, opened 2021-01-27
Connections/debug and improvements

pv3nturi:
- DebugFilter to add come connections data with  endpoints.

- When release forceClose is performed now connection is returned only when closeFuture complete.


carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
```diff
@@ -393,14 +394,19 @@ public void release(boolean close, RequestHandler clientSidePeerHandler, Runnabl
...
-    void destroy() {
+    void destroy(Runnable onCloseCompleted) {
         invalidate();
         if (!closed.compareAndSet(false, true)) {
             return;
 
         }
         // note: close is async, we are not waiting for real close
-        channelToEndpoint.close();
+        ChannelFuture closeFuture = channelToEndpoint.close();
```
nicoloboschi, line 404: If we want to wait we have to call .get() after the close()

carapace-server/src/main/java/org/carapaceproxy/server/filters/DebugFilter.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Adds a debug header.
+ *
+ * @author paolo.venturi
+ */
+public class DebugFilter extends BasicRequestFilter {
+
+    public static final String TYPE = "debug";
+    public static final String DEBUG_HEADER_NAME = "X-Carapace-Debug";
```
nicoloboschi, line 35: X-Carapace-CID or similar

nicoloboschi (changes requested):
It looks good, I left two comments

carapace-server/src/main/java/org/carapaceproxy/server/filters/DebugFilter.java
pv3nturi: I meant that as a list of debug info for future improvements (and low headers spam)

pv3nturi:
Let's create a brand new branch with debug-header only.

PR #247, opened 2021-01-29
Requests header debug

pv3nturi:
Debug header added to keep track of connections id from carapace to endpoints


nicoloboschi (approved):
+1

PR #249, opened 2021-02-01
Configuration property for clients idle timeout

pv3nturi:
New config property for clients idle timeout: _**clients.idle.timeout**_

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
@@ -27,6 +27,8 @@ import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.handler.ssl.SslHandler;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Gauge;
 import java.net.SocketAddress;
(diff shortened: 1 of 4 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
```diff
@@ -132,6 +134,17 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
         LOG.log(Level.SEVERE, "bad error", cause);
     }
 
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+        if (evt instanceof IdleStateEvent) {
+            IdleStateEvent e = (IdleStateEvent) evt;
+            if (e.state() == IdleState.ALL_IDLE) {
+                LOG.log(Level.SEVERE, "Idle connection detected for client {0}. Channel closed.", new Object[]{clientAddress});
```
hamadodene, line 142: Maybe debug?

hamadodene (approved):
LGTM

carapace-server/src/test/java/org/carapaceproxy/RawClientTest.java
```diff
@@ -565,6 +565,49 @@ public void testRequestsReadTimeout() throws Exception {
...
+            RuntimeServerConfiguration currentConfiguration = server.getCurrentConfiguration();
+            currentConfiguration.setClientsIdleTimeoutSeconds(30);
+            server.getConnectionsManager().applyNewConfiguration(currentConfiguration);
+
+            stats = server.getConnectionsManager().getStats();
+
+            try (RawHttpClient client = new RawHttpClient("localhost", port)) {
+                for (int j = 0; j < 2; j++) {
+                    RawHttpClient.HttpResponse res = client.get("/index.html");
+                    Thread.sleep(60_000);
```
nicoloboschi, line 595: maybe it is better to change the new parameter with a lower value and reduce test duration?

carapace-server/src/main/java/org/carapaceproxy/server/ClientConnectionHandler.java
nicoloboschi: +1 for debug level

nicoloboschi (changes requested):

nicoloboschi (approved):

Issue #251, opened 2021-02-04
Web page doesn't work in HTTPS during certificate renewal


hamadodene:
```
183855291-Feb 04, 2021 3:34:16 PM org.carapaceproxy.server.certificates.DynamicCertificatesManager getCertificateForDomain
2183855404:SEVERE: Dynamic certificate for domain *.mydomian is not available: CertificateData{domain=*.mydomian , state=ordering, available=false, manual=false}
```

While renewing a certificate, we noticed the web pages no longer respond correctly in https.
The correct behavior should be that until the certificate has been renewed, the latest available certificate is use.

pv3nturi:
We need to drop the isAvailable flag.. Whether the certificate is expired we has to serve it as it is. Fallback certificate has to be served ONLY whether the certificate does not exists.

hamadodene:
+1 for drop isAvaible flag
We should have something like:

- If the certificate is being renewed, serve the latest available certificate
- if the certificate has expired, serve it as it is
- if the certificate is not configured, fallback to the default certificate
- If the certificate is static and not working, fallback to the default certificate
- if the static certificate has expired, serve it as is

PR #256, opened 2021-02-22
Web page doesn't work in HTTPS during certificate renewal #251

pv3nturi:
PR for #251 

Changelog:

- empty certificates chains will fallback to default certificate
- removed "available" flag for dynamic certificates

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
@@ -1,21 +1,21 @@
 /*
- Licensed to Diennea S.r.l. under one
- or more contributor license agreements. See the NOTICE file
- distributed with this work for additional information
- regarding copyright ownership. Diennea S.r.l. licenses this file
- to you under the Apache License, Version 2.0 (the
- "License"); you may not use this file except in compliance
- with the License.  You may obtain a copy of the License at
-
- http://www.apache.org/licenses/LICENSE-2.0
-
- Unless required by applicable law or agreed to in writing,
- software distributed under the License is distributed on an
- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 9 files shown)


carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
```diff
@@ -85,12 +85,12 @@
     private static final String DIGITAL_CERTIFICATES_TABLE_NAME = "digital_certificates";
     private static final String CREATE_DIGITAL_CERTIFICATES_TABLE = "CREATE TABLE " + DIGITAL_CERTIFICATES_TABLE_NAME
             + "(domain string primary key, privateKey string, chain string, "
-            + "state string, pendingOrder string, pendingChallenge string, available tinyint)";
+            + "state string, pendingOrder string, pendingChallenge string)";
```
hamadodene, line 88: You have deleted a field from the table. For those who already have that table full, it is necessary to make that field nullable first. Otherwise the insertion fails

hamadodene (changes requested):

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
hamadodene: @pv3nturi I just did a manual check  and by default the column are created nullable. 
```
INSERT INTO  digital_certificates (domain, privateKey, chain, state, pendingOrder, pendingChallenge) values ('example.com', 'test', 'test', 'test', 'test', 'test')
```

```
| domain   | privatekey | chain | state | pendingorder | pendingchallenge | available |
| ------   | ---------- | ----- | ----- | ------------ | ---------------- | --------- |
| example.com | test| test| test| test| test| <NULL>    |
```

hamadodene (approved):

nicoloboschi (approved):
LGTM

Issue #260, opened 2021-02-23
Manage cache mode in HTTPS requests


hamadodene:
If the call is in HTTPS it should not be cached by the intermediate (eg. the Reverse, such as Carapace), unless there is cache-control=public. If there is no Cache-Control = public it can be treated as a Cache-Control = private.

aluccaroni:
NB: we don't need to change the content of the cache-control header...only evaluate if cache-control=public is present if the call is in https before caching

PR #263, opened 2021-02-23
Manage cache mode in HTTPS requests #260

pv3nturi:
PR for #260

Changelog:
- Https requests won't be cached anymore whether header _Cache-Control_ != _public_
   - this is achieved only whether dynamic property **_cache.requests.secure.disablewithoutpublic = true_** is set
- Http/Https requests won't be cached anymore whether
  - header _Pragma: no-cache_ is present
  - header _Cache-Control_ has values: _private_, _no-cache_, _no-store_ or _max-age=0_

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java
@@ -78,8 +78,11 @@ public class ContentsCache {
         this.stats = new CacheStats();
         this.threadPool = Executors.newSingleThreadScheduledExecutor();
 
-        this.currentConfiguration = new CacheRuntimeConfiguration(currentConfiguration.getCacheMaxSize(), currentConfiguration.getCacheMaxFileSize());
-
+        this.currentConfiguration = new CacheRuntimeConfiguration(
+                currentConfiguration.getCacheMaxSize(),
+                currentConfiguration.getCacheMaxFileSize(),
+                currentConfiguration.isCacheDisabledForSecureRequestsWithoutPublic()
+        );
         this.cache = new CaffeineCacheImpl(stats, currentConfiguration.getCacheMaxSize(), LOG);
     }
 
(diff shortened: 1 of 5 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java
```diff
@@ -154,14 +154,22 @@ private boolean isCachable(HttpRequest request, boolean registerNoCacheStat) {
...
+        final String cacheControl = headers.get(HttpHeaderNames.CACHE_CONTROL, "").replaceAll(" ", "").toLowerCase();
+        boolean ctrlF5 = cacheControl.contains(HttpHeaderValues.NO_CACHE);
         if (ctrlF5) {
             if (registerNoCacheStat) {
                 NO_CACHE_REQUESTS_COUNTER.inc();
             }
             return false;
         }
+        if ((secure && !cacheControl.contains(HttpHeaderValues.PUBLIC)
+                || headers.contains(HttpHeaderNames.PRAGMA, HttpHeaderValues.NO_CACHE, true))) {
```
hamadodene, line 168: please handle case sensitive, and when we have header with multiple part like Cache-Control: public, max-age 3600

hamadodene (changes requested):

carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java
pv3nturi: This is how actually works

hamadodene: ok look good

hamadodene (approved):

hamadodene (approved):

aluccaroni:
Since this "feature" is not linked to any existing RFC we need to add a property to enable it and it should be disabled by default

nicoloboschi (approved):
+1

PR #266, opened 2021-02-24
Perform dns check before create/renew certificate #239

pv3nturi:
PR for #239 

Changelog:

- domain dns checking before acme certificates order creation
- _**dynamiccertificatesmanager.domainsreachabilitychecker.ipaddresses**_ to define carapace ips that acme certificate domains has to point to.

Issue #239, opened 2021-01-22
Perform dns check before create/renew certificate


hamadodene:
It would be better to do a dns check (to check if the DNS points to an IP used by the system) before ACME certificates creation or renewal.

This is mainly to avoid problems related to rate limiting (too much request to the certificate provider)

PR #270, opened 2021-03-15
Perform dns check before create/renew certificate #239 #266

pv3nturi:
New dynamic certificates state: **_domain unreachable_**


carapace-server/src/test/java/org/carapaceproxy/ConnectionPoolTest.java
```diff
@@ -58,8 +58,6 @@
 
 public class ConnectionPoolTest {
 
-    public static final long KEEP_ALIVE_TIMEOUT = 30_000;
```
pv3nturi, line 61: Did not work

nicoloboschi (approved):
+1

PR #275, opened 2021-03-17
Perform dns check before create/renew certificate #239

pv3nturi:
IP check has not to be performed for wildcard certificates


nicoloboschi (approved):
+1

PR #277, opened 2021-03-18
Advanced logger: disable fireChannelRead when disabled

pv3nturi:
Whether the advanced logger is disabled we has not to call ctx.fireChannelRead


nicoloboschi (approved):
+1

PR #281, opened 2021-04-29
Improved server-client connection release + ssl chain cache

pv3nturi:
Changelog:

- whether the server send connection close now we close the channel to the client
- release connection to endpoint with forceClose flag now skip client check for release
- KeyManagerFactory is now wrapped by OpenSslCachingX509KeyManagerFactory (this should improve performance during ssl connections establishment)


nicoloboschi (approved):
+1

PR #282, opened 2021-05-05
In-memory certificates data improvement

pv3nturi:
Improved certificates data management on memory:

- serial number, expiring date and keystore data are now keep in memory and not more retrived from decoded chain

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/certificates/DynamicCertificatesManager.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/certificates/DynamicCertificatesManager.java
@@ -19,7 +19,6 @@
  */
 package org.carapaceproxy.server.certificates;
 
-import static org.carapaceproxy.configstore.ConfigurationStoreUtils.base64DecodeCertificateChain;
 import static org.carapaceproxy.configstore.ConfigurationStoreUtils.base64EncodeCertificateChain;
 import static org.carapaceproxy.server.certificates.DynamicCertificateState.AVAILABLE;
 import static org.carapaceproxy.server.certificates.DynamicCertificateState.DNS_CHALLENGE_WAIT;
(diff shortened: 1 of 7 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/certificates/DynamicCertificatesManager.java
```diff
@@ -273,11 +284,10 @@ private void certificatesLifecycle() {
                 .sorted((e1, e2) -> e1.getKey().compareTo(e2.getKey()))
                 .map(e -> e.getValue())
                 .collect(Collectors.toList());
-        for (CertificateData data : _certificates) {
+        for (CertificateData cert : _certificates) {
             boolean updateCertificate = true;
-            final String domain = data.getDomain();
+            final String domain = cert.getDomain();
             try {
-                CertificateData cert = loadOrCreateDynamicCertificateForDomain(domain, data.isWildcard(), false, data.getDaysBeforeRenewal());
```
pv3nturi, line 280: We skip retriving the certificate from db which is already up-to-date. (all certs are reloaded as soon as a node triggers fire-cache request)

carapace-server/src/main/java/org/carapaceproxy/utils/CertificatesUtils.java
```diff
@@ -146,19 +146,14 @@ public static KeyStore loadKeyStoreData(byte[] data, String password)
...
-        if (chain == null || chain.length == 0) {
+    public static boolean isCertificateExpired(Date expiringDate, int daysBeforeRenewal) throws GeneralSecurityException {
+        if (expiringDate == null) {
             return false;
         }
-        try {
-            Calendar cal = Calendar.getInstance();
-            cal.setTime(new Date());
-            cal.add(Calendar.DATE, daysBeforeRenewal);
-            ((X509Certificate) chain[0]).checkValidity(cal.getTime());
```
pv3nturi, line 157: This checks notBefore and notAfter. Maybe we just need notAfter.

nicoloboschi (changes requested):
It looks good, I left a comment on the expire date check

nicoloboschi (approved):
+1

PR #292, opened 2021-06-04
Fix static certificates cache

pv3nturi:
Added static certificates keystores cache


pv3nturi:
Certificates are still cached by listener_host:port+snihostname

PR #295, opened 2021-06-07
Listeners > BootSSLContext: improve locking access to cached sslConte…

pv3nturi:
PR for #294

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/Listeners.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/Listeners.java
@@ -73,6 +73,7 @@ import io.netty.handler.timeout.IdleStateHandler;
 import static org.carapaceproxy.utils.CertificatesUtils.loadKeyStoreFromFile;
 import io.netty.handler.ssl.OpenSslCachingX509KeyManagerFactory;
 import io.netty.util.concurrent.Future;
+import java.util.HashMap;
 
 /**
  *
@@ -277,7 +278,7 @@ public class Listeners {
 
     private final class ListenerSniMapping implements io.netty.util.AsyncMapping<String, SslContext> {
 
-        private final Map<String, SslContext> listenerSslContexts = new ConcurrentHashMap<>();
+        private final Map<String, SslContext> listenerSslContexts = new HashMap<>();
         private final NetworkListenerConfiguration listener;
         private final int port;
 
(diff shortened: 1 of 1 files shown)


carapace-server/src/main/java/org/carapaceproxy/server/Listeners.java
```diff
@@ -282,33 +275,59 @@ public void stop() {
...
     }
 
-    private SslContext resolveSslContext(NetworkListenerConfiguration listener, String sniHostname) throws ConfigurationNotValidException {
-        int port = listener.getPort() + parent.getListenersOffsetPort();
-        String key = listener.getHost() + ":" + port + "+" + sniHostname;
-        if (LOG.isLoggable(Level.FINER)) {
-            LOG.log(Level.FINER, "resolve SNI mapping " + sniHostname + ", key: " + key);
+    private final class ListenerSniMapping implements io.netty.util.AsyncMapping<String, SslContext> {
+
+        private final Map<String, SslContext> listenerSslContexts = new ConcurrentHashMap<>();
```
nicoloboschi, line 280: no need to be Concurrent as is accessed only by this thread

carapace-server/src/main/java/org/carapaceproxy/server/Listeners.java
```diff
@@ -282,33 +275,59 @@ public void stop() {
...
                     SSLCertificateConfiguration choosen = chooseCertificate(sniHostname, listener.getDefaultCertificate());
                     if (choosen == null) {
                         throw new ConfigurationNotValidException("cannot find a certificate for snihostname " + sniHostname
                                 + ", with default cert for listener as '" + listener.getDefaultCertificate()
                                 + "', available " + currentConfiguration.getCertificates().keySet());
                     }
-                    return bootSslContext(listener, choosen);
-                } catch (ConfigurationNotValidException err) {
-                    throw new RuntimeException(err);
+                    sslContext = bootSslContext(listener, choosen);
```
nicoloboschi, line 314: idk if you did it on purpose, but now there is a possibility where more than one listener will create the SslContext for the same key
btw this is a less locking-situation so I hope this will decrease latency (maybe with more CPU usages in some cases)

carapace-server/src/main/java/org/carapaceproxy/server/Listeners.java
```diff
@@ -282,33 +275,59 @@ public void stop() {
...
+                    listenerSslContexts.put(key, sslContext);
+                    sslContexts.put(key, sslContext);
```
nicoloboschi, line 316: sslContexts is invalidated when the config changes, we need to invalidate this local map as well

nicoloboschi (commented):
It is promising, but you are missing local map invalidation

carapace-server/src/main/java/org/carapaceproxy/server/Listeners.java
pv3nturi: whether config changes all listeners are recreated so the local mapping

nicoloboschi (approved):
+1

PR #297, opened 2021-06-16
Bug/100 continue management

pv3nturi:
Whenever a client makes a request including header "Expect: 100-continue" the server sends back a 100 continue response together with a LastHttpContent. This breaks the "client-server release handshake" so that another client may use the same connection whenever the first client (c1) sends its LastHttpContent (performing delayed release recovery) and getting the response  of c1.

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
@@ -457,7 +457,8 @@ public class EndpointConnectionImpl implements EndpointConnection {
         public void channelRead0(ChannelHandlerContext ctx, HttpObject msg) {
             RequestHandler _clientSidePeerHandler = clientSidePeerHandler.get();
             if (_clientSidePeerHandler == null || !requestRunning) {
-                LOG.log(Level.INFO, "swallow content {0}: {1}, disconnected client. connection: {2}", new Object[]{msg.getClass(), msg, EndpointConnectionImpl.this});
+                final String cause = _clientSidePeerHandler == null ? "no more client connected" : "request stopped";
+                LOG.log(Level.INFO, id + ": swallow content {0}: {1}, disconnected client due to {2}. connection: {3}", new Object[]{msg.getClass(), msg, cause, EndpointConnectionImpl.this});
                 return;
             }
             if (msg instanceof HttpContent) {
@@ -479,7 +480,7 @@ public class EndpointConnectionImpl implements EndpointConnection {
         public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
             RequestHandler _clientSidePeerHandler = clientSidePeerHandler.get();
             if (_clientSidePeerHandler != null && requestRunning) {
-                logConnectionInfo("channelReadComplete, open: " + ctx.channel().isOpen(), "");
+                logConnectionInfo("channelReadComplete, open: " + ctx.channel().isOpen());
                 _clientSidePeerHandler.readCompletedFromRemote();
             }
         }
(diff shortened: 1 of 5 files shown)


carapace-server/src/main/java/org/carapaceproxy/client/impl/EndpointConnectionImpl.java
```diff
@@ -534,8 +534,16 @@ public void setRequestsHeaderDebugEnabled(boolean requestsHeaderDebugEnabled) {
...
     }
 
-    private void logConnectionInfo(String s) {
-        CarapaceLogger.debug("{0}: {1}", s, EndpointConnectionImpl.this);
+    private void logConnectionInfo(String desc) {
+        logConnectionInfo(desc, "");
+    }
+
+    private void logConnectionInfo(String desc, Object info) {
+        if (info != null && info instanceof HttpContent) {
```
nicoloboschi, line 542: this should be done only if debug is enabled, it's really expensive

carapace-server/src/test/java/org/carapaceproxy/RawClientTest.java
```diff
@@ -750,4 +789,216 @@ public void testClientConnectionClose() throws Exception {
...
+                try {
+                    futures.add(ex.submit(() -> {
+                        try (RawHttpClient client1 = new RawHttpClient("localhost", port)) {
+                            String body = "filler-content";
+                            String request = "POST /index.html HTTP/1.1"
+                                    + "\r\nHost: localhost"
+                                    + "\r\nConnection: keep-alive"
+                                    + "\r\nContent-Type: text/plain"
+                                    + "\r\n" + HttpHeaderNames.EXPECT + ": " + HttpHeaderValues.CONTINUE
+                                    + "\r\ntrigger: continue"
```
nicoloboschi, line 832: Should we use standard header Expect to verify 100-continue beehaviour

carapace-server/src/test/java/org/carapaceproxy/RawClientTest.java
```diff
@@ -750,4 +789,216 @@ public void testClientConnectionClose() throws Exception {
...
+                            Thread.sleep(1_000);
+
+                            oo.write(body.getBytes(StandardCharsets.UTF_8));
+                            oo.flush();
+
+                            String resp = consumeHttpResponseInput(socket.getInputStream()).getBodyString();
+                            System.out.println("RESP client1: " + resp);
+                            if (!resp.contains("HTTP/1.1 100 Continue")) {
+                                failed.set(true);
+                            }
```
nicoloboschi, line 850: ok, then we should expect the final response, as a real-world http client

carapace-server/src/test/java/org/carapaceproxy/RawClientTest.java
```diff
@@ -750,4 +789,216 @@ public void testClientConnectionClose() throws Exception {
...
+//                                failed.set(true);
+//                            }
+                        } catch (Exception e) {
+                            System.out.println("EXCEPTION: " + e);
+                            failed.set(true);
+                        }
+                    }));
+                    futures.add(ex.submit(() -> {
+                        try {
+                            Thread.sleep(10_000);
```
nicoloboschi, line 863: can we reduce/remove this long sleep using code hooks to manage events order?

carapace-server/src/test/java/org/carapaceproxy/utils/RawHttpClient.java
```diff
@@ -45,7 +45,7 @@
  */
 public final class RawHttpClient implements AutoCloseable {
 
-    private static boolean DEBUG = false;
+    private static boolean DEBUG = true;
```
nicoloboschi, line 48: we should leave false

nicoloboschi (changes requested):
+1, but test can be improved

nicoloboschi (approved):
+1

PR #298, opened 2021-06-21
Improve client-endpoints connection management

pv3nturi:
- Improved connections exclusively reuse
- Endpoints connections released right after clients one is closed (client-endpoint release handshake dropped)
- Requests form same client (re)uses same endpoint connection
- Clients endpoint-connection reuse reset on config reload and endpoint down

PR #300, opened 2021-07-01
Revert/improve client endopoint connection management

pv3nturi:
Revert of Improve client-endpoints connection management #298


pv3nturi:
This pr is no more useful due to Reactor netty migration #303

PR #316, opened 2022-01-19
bumping core libs

pv3nturi:
Bumping core libs (no tests) to latest version before v1.7 release


aluccaroni:
@pv3nturi regarding jetty we could update to 9.4.44.v20210927

hamadodene (approved):

PR #341, opened 2022-04-11
Fix direct memory leak

pv3nturi:
Fix to ByteBuf releasing


carapace-server/src/main/java/org/carapaceproxy/server/cache/CaffeineCacheImpl.java
```diff
@@ -159,6 +159,8 @@ public CachedContent get(ContentKey key) {
     }
 
     private void release(ContentKey key, CachedContent payload) {
+        logger.log(Level.FINE, "Releasing cache content {0}", new Object[]{key.uri});
```
dmercuriali, line 162: wrap this in a `if (LOG.isLoggable(Level.FINE)) {` like below

dmercuriali (changes requested):

dmercuriali (approved):

PR #342, opened 2022-04-20
Fix micrometrics empty

pv3ntur1:
Fix to micrometrics empty after config reload


hamadodene (approved):
I have test this patch..good to me.

dmercuriali (approved):

PR #349, opened 2022-05-17
Debub tomcat request from proxy to backend

hamadodene:


pv3nturi (approved):
A unique method should be better instead of copying&pasting everywhere the log

PR #353, opened 2022-05-19
Improving requests forwarding

pv3nturi:
In order to avoid proxy-to-server-clients setup at every proxy request now we use an existent client from a pool configured by endpoint-connectionpool.


hamadodene (approved):
Good to me.
@dmercuriali  can you review?

PR #366, opened 2022-07-22
Hide spamming logs for unsupported channel options

pv3nturi:
In production we see a lot of spammy log at every client request


hamadodene (approved):
ok to me

PR #372, opened 2022-08-23
Carapace > Provide maintenance mode

hamadodene:
- provide dynamic config - **carapace.maintenancemode.enabled=true/false**
- added possibility of configuring a maintenance page per route (ex. **route.1.maintenanceaction=maintenance-static-action**)

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/core/HttpProxyServer.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/core/HttpProxyServer.java
@@ -634,13 +634,13 @@ public class HttpProxyServer implements AutoCloseable {
         Properties props = dynamicConfigurationStore.asProperties(null);
         boolean newCertificate = !dynamicConfigurationStore.anyPropertyMatches((k, v) -> {
             if (k.matches("certificate\\.[0-9]+\\.hostname") && v.equals(cert.getDomain())) {
-                performUpdate(props, k, cert); // updating existing certificate
+                performCertificateUpdate(props, k, cert); // updating existing certificate
                 return true;
             }
             return false;
         });
         if (newCertificate) {
-            performCreate(props, cert);
+            performCertificateCreate(props, cert);
         }
 
         // Configuration reloading
@@ -653,7 +653,7 @@ public class HttpProxyServer implements AutoCloseable {
         applyDynamicConfigurationFromAPI(new PropertiesConfigurationStore(props));
     }
 
-    private void performUpdate(Properties props, String key, CertificateData cert) {
+    private void performCertificateUpdate(Properties props, String key, CertificateData cert) {
         props.setProperty(key.replace("hostname", "mode"), cert.isManual() ? "manual" : "acme");
         if (cert.isManual()) {
             props.remove(key.replace("hostname", "daysbeforerenewal")); // type changed from acme to manual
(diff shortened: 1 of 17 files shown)


hamadodene:
Schermata 2022-08-24 alle 10 26 40

hamadodene:
you should put a warning in the navbar, or rather change the color of the navbar with a message.
In my opinion, better to do it in another issue. @pv3nturi Can you take care of it?

hamadodene:
@pv3nturi PTAL
I added test also for api

carapace-server/src/main/java/org/carapaceproxy/core/HttpProxyServer.java
```diff
@@ -717,6 +723,10 @@ private void applyDynamicConfiguration(ConfigurationStore newConfigurationStore,
                 proxyRequestsManager.reloadConfiguration(newConfiguration, newBackends.values());
             }
 
+            if(currentConfiguration.isMaintenanceModeEnabled() != newConfiguration.isMaintenanceModeEnabled()) {
```
pv3nturi, line 726: Is this necessary? applyDynamicConfigurationFromAPI should do this

hamadodene: I think it is necessary, because otherwise the releoad of the proxyRequestManager is done only when the backend configuration changes.

The only place where maintenance is checked is in the proxyRequestManager. So we have to reload even when the maintenance status changes.

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
```diff
@@ -125,6 +128,12 @@ public Publisher<Void> processRequest(ProxyRequest request) {
             LOGGER.log(Level.INFO, "Mapper returned NULL action for {0}", this);
             action = MapResult.internalError(MapResult.NO_ROUTE);
         }
+
+        if(currentConfiguration.isMaintenanceModeEnabled()) {
```
pv3nturi, line 132: This should be done by StandardEndpointMapper

pv3nturi (changes requested):

hamadodene:
Schermata 2022-09-21 alle 19 03 17

carapace-server/src/main/java/org/carapaceproxy/core/HttpProxyServer.java
pv3nturi, line 726: I think this is useless, because this code is run only when a certificare is uploaded from api and in thad case the mantainance mode shouldn't change

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
```diff
@@ -110,6 +109,10 @@ public void reloadConfiguration(RuntimeServerConfiguration newConfiguration, Col
         connectionsManager.reloadConfiguration(newConfiguration, newEndpoints);
     }
 
+    public void reloadConfiguration(RuntimeServerConfiguration newConfiguration) {
```
pv3nturi, line 112: this can be dropped, is no more usefull

pv3nturi (changes requested):
Leave some comments

hamadodene:
@pv3nturi Fixed your comments, PTAL.

pv3nturi (approved):

PR #378, opened 2022-10-12
Avoid using chunked transfer encoding while replying to HTTP/1.0 requests

hamadodene:
Since HTTP 1.0 clients aren't required to support chunked transfer encoding, it shouldn't be used upon responding to HTTP/1.0 requests. Otherwise, a non-supporting client would just ignore the encoding and deliver a messed up response body to the application layer.

Implementation:
provide a dynamic configuration to enable or disabile this fix:

**carapace.http10backwardcompatibility.enabled=true/false** (default false)

More info:
[Writing data reactor netty](https://projectreactor.io/docs/netty/release/reference/index.html#_writing_data_4)
[Reactor netty set content lenght automatically](https://stackoverflow.com/questions/57802358/netty-set-content-length-automatically)

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
@@ -19,58 +19,52 @@
  */
 package org.carapaceproxy.core;
 
-import static org.carapaceproxy.server.mapper.MapResult.*;
-import static reactor.netty.Metrics.CONNECTION_PROVIDER_PREFIX;
 import com.google.common.annotations.VisibleForTesting;
 import io.micrometer.core.instrument.Metrics;
+import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.epoll.Epoll;
 import io.netty.channel.epoll.EpollChannelOption;
 import io.netty.channel.socket.nio.NioChannelOption;
-import io.netty.handler.codec.http.DefaultFullHttpResponse;
-import io.netty.handler.codec.http.DefaultHttpHeaders;
-import io.netty.handler.codec.http.FullHttpResponse;
-import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpHeaderValues;
-import io.netty.handler.codec.http.HttpHeaders;
-import io.netty.handler.codec.http.HttpResponseStatus;
-import io.netty.handler.codec.http.HttpVersion;
+import io.netty.handler.codec.http.*;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Gauge;
-import java.net.ConnectException;
-import java.net.InetSocketAddress;
-import java.time.Duration;
-import java.time.LocalDateTime;
-import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.Function;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import java.util.regex.Pattern;
 import jdk.net.ExtendedSocketOptions;
 import org.carapaceproxy.EndpointStats;
 import org.carapaceproxy.SimpleHTTPResponse;
-import org.carapaceproxy.server.mapper.MapResult;
 import org.carapaceproxy.client.EndpointKey;
 import org.carapaceproxy.server.cache.ContentsCache;
 import org.carapaceproxy.server.config.BackendConfiguration;
 import org.carapaceproxy.server.config.ConnectionPoolConfiguration;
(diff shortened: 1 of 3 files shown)


carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
```diff
@@ -428,6 +433,22 @@ public Publisher<Void> forward(ProxyRequest request, boolean cache) {
                     }
                     addCustomResponseHeaders(request, request.getAction().customHeaders);
 
+                    if (currentConfiguration.isHttp10BackwardCompatibilityEnabled() &&
```
pv3nturi, line 436: Maybe it's better to call parent.getCurrentConfiguration().isHttp10BackwardCompatibilityEnabled(), this way we don't need to call reloadConfiguration which destroys all connections pool

carapace-server/src/main/java/org/carapaceproxy/core/HttpProxyServer.java
```diff
@@ -719,7 +719,8 @@ private void applyDynamicConfiguration(ConfigurationStore newConfigurationStore,
             Map<String, BackendConfiguration> newBackends = newMapper.getBackends();
             this.mapper = newMapper;
 
-            if (!newBackends.equals(currentBackends) || isConnectionsConfigurationChanged(newConfiguration)) {
+            if (!newBackends.equals(currentBackends) || isConnectionsConfigurationChanged(newConfiguration)
+                    || currentConfiguration.isHttp10BackwardCompatibilityEnabled() != newConfiguration.isHttp10BackwardCompatibilityEnabled()) {
```
pv3nturi, line 723: We can avoid this, by just using parent.getCurrentConfiguration()

pv3nturi (changes requested):
Unit test left

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
hamadodene: @pv3nturi Right, make sense

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
```diff
@@ -522,6 +532,11 @@ private Publisher<Void> serveFromCache(ProxyRequest request, ContentsCache.Conte
             headers.add(HttpHeaderNames.EXPIRES, HttpUtils.formatDateHeader(new java.util.Date(content.getExpiresTs())));
             request.setResponseHeaders(headers);
             addCustomResponseHeaders(request, request.getAction().customHeaders);
+            //If the request is http 1.0 we make sure to send without chunked
+            if (parent.getCurrentConfiguration().isHttp10BackwardCompatibilityEnabled() &&
+                    request.getRequest().version() == HttpVersion.HTTP_1_0) {
```
hamadodene, line 537: @pv3nturi We need to introduce this part, becuse Flux.fromIterable return a new Flux and publisher in this case can emit more than 1 element. 

However Mono.from ensure that publisher will emit 0 or 1 item.

pv3nturi: Good catch!

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
```diff
@@ -522,6 +532,11 @@ private Publisher<Void> serveFromCache(ProxyRequest request, ContentsCache.Conte
...
+                return request.sendResponseData(Mono.from(ByteBufFlux.fromIterable(content.getChunks())));
```
pv3nturi, line 538: Is it possible to create the Mono directly from content.getChuks withot using ByteBufFlux.fromIterable?

carapace-server/src/test/java/org/carapaceproxy/backends/ChunkedEncodingResponseTest.java
```diff
@@ -139,4 +142,73 @@ public void testSimpleChunckedResponseWithCache() throws Exception {
...
+            server.getCache().getStats().resetCacheMetrics();
+            server.getCache().clear();
+            assertEquals(0, server.getCache().getCacheSize());
+            assertEquals(0, server.getCache().getStats().getHits());
+
+            request.setProtocolVersion(HttpVersion.HTTP_1_0);
+            httpresponse = httpclient.execute(request);
+
+            assertNull(httpresponse.getFirstHeader(String.valueOf(HttpHeaderNames.TRANSFER_ENCODING)));
+            assertNotNull(httpresponse.getFirstHeader(String.valueOf(HttpHeaderNames.CONTENT_LENGTH)));
```
pv3nturi, line 183: Maybe we can assert that the response was not from cache

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequestsManager.java
hamadodene: @pv3nturi  It's not possibile. content.getChunks is a List of ByteBuf and Mono doesn't have any function to convert a List of Bytebuf

pv3nturi (approved):

Issue #324, opened 2022-02-22
Certificate > sync certificates locally


hamadodene:
Currently to have certificate file generated by Carapace, you have to manually using API.
For internal use, we would need possibility to sync the certificates automatically on a local path:
- If a new certificate is created, it must also be written locally in its local folder. Ex. (/Domain_name/domain.p12)
- If a certificate is renewed, the local certificate must be updated with the new one
- it must be possible to specify the folder where to save all the certificates. Within the specified folder, each certificate must still have its own folder.

aluccaroni:
@hamadodene  we need to decide how to handle the case when the directory is not writable and/or the write fail. Also, it should be possible to manually force carapace to write down the cert (eg. if it fail/it's lost we need a way to dave it down again)

hamadodene:
@pv3nturi  @aluccaroni 
 I think we can provide two API to handle write fail problem:
- An API to force Carapace to write down all dynamic certificate  (This can be dangerous if there are a lot of certificates. If we add a button in Admin interface, we need a visible warning.)
- An API to force Carapace to write down an specific certificate

For both API we can provide an button in ADMIN interface.

Also, if the user provides an unwritable path, then we don't have to allow the configuration to be saved.
@pv3nturi  So we can verify that the folder is writable in the configuration validation.

PR #379, opened 2022-10-17
Certificate > sync certificates locally #324

pv3nturi:
Addes these dynamic props:

- dynamiccertificatesmanager.localcertificates.store.path: to enable the local store of acme certificates to specified path

- dynamiccertificatesmanager.localcertificates.peers.ids: to enable the local store for specific peers (default is enabled for all pees)

All acme certificates are dumped in local storage when became  AVAILABLE or when forced from admin (by dumping a specified cert or all of them).

Added support to attach data to zookeeper events.
Also, the column privateKey of digital_certificates table has been dropped.
Finally a fix to maintenance mode , buttons and sidebar styles


hamadodene (approved):
LGTM
i tested this branch and all is ok.

Issue #380, opened 2022-10-24
Possibility to generate a certificate that covers multiple domains


hamadodene:
In some cases it may be necessary to generate a certificate that covers several domains such as:
**example.com www.example.com**
or even a certificate covering wildcard and domain itself like: **example.com * .example.com**

Another example:
- www.example.com www.example.it

It would be useful to enable this possibility.

aluccaroni:
@hamadodene we need to support the Subject Alternative Name extensions. I suggest to add a new configuration line, something like:
certificate.ID.hostname=example.com
certificate.ID.san=www.example.com,www2.example.com
certificate.ID.mode=acme

hamadodene:
I agree with @aluccaroni 
There would be the case of multiple wildcard certificates.
In the Letsencrypt forum they say it's possible. I would first evaluate how complex it is to do it with acme4j (See https://community.letsencrypt.org/t/multi-domain-wildcard-certificate/113577/9). If too complex we can avoid support with multiple wildcard certificates. 

By multiple wildcard certificates I mean something like:

certificate.ID.hostname=*.example.com
certificate.ID.san=*.example2.com 
certificate.ID.mode=acme

PR #386, opened 2022-11-17
Possibility to generate a certificate that covers multiple domains #380

pv3nturi:
### Dynamic configuration
 Added the new property 'san' to set a list of subject alternative names for a certificate:
`certificate.1.san=example.com,example2.com`

### Database
Added the new string columns 'subjectAltNames' and 'pendingChallenges' in table digital_certificates

### Certificate ordering
The whole certificate ordering lifecycle has been changed in order to manage multiple pending challenges for the same order:

- When a certificate with one or more subject alternative name is ordered, for each san is created and verified a challenge html (for plain domains) or dns (for wildcard domains).

- Only if all challenges succeeded the order is verified; with a single challenge failed the whole order is discarded. 

- As soon as all challenges have been verified and the order performed, the resulting certificate includes both the main domain and the sans
 
### Certificate request matching
Whenever a request is performed the certificate is chosen according to the main domain and the subject alternative names.


dmercuriali (changes requested):
Most of this PR changes has nothing to do with the actual issue. We should create another issue and move the refactors/reformats there, or at least move the actual implementation in another commit.

From what i can see the implementation seems fine, but I would change the var/column name to be more explicit

PR #392, opened 2022-12-23
Certificates attempts and renewal

### LINKED ISSUE #264: Dynamic certificates > Manage the maximum number of attempts when renewal or certificate creation fails
```
                    case REQUEST_FAILED: { // challenge/order failed
                        LOG.log(Level.INFO, "Certificate issuing for domain: {0} current status is FAILED, setting status=WAITING again.", domain);
                        cert.setState(WAITING);
                        break;
                    }

```
In the management of dynamic certificates, I have seen that when we are in the case of request_failed we put the certificate back in the waiting state. So let's try renewal forever. Knowing the limits of letecnryt for errors, in my opinion it is better to have a max-retry after which to put the certificate in the final failed state.
We can then have an api to unlock it (also linked to a button on the GUI).

https://letsencrypt.org/docs/rate-limits/

### LINKED ISSUE #359: Certificate > Report the error on the interface in case of failure
It has happened in some cases that the generation of a certificate fails for some reason. It would be advisable to report the failure on the interface and also on the logs.

### LINKED ISSUE #401: Dynamic certificate > Domain unreachable if no SAN is specified
Doing some checks I realized that if no SAN is specified, the certificate creation fails with DOMAIN UNREACHBLE.
From code I noticed that subjectAltNames is a Set<String> and when we go to create the collection with all the certificates we verify that this is not null.

```
Class: DynamicCertificatesManager

    final var allNamesReachable = cert.getNames().stream().allMatch(name -> {
        try {
            return CertificatesUtils.isWildcard(name) || checkDomain(name);
        } catch (AcmeException e) {
            return false;
        }
    });
```
```
Class  CertificateData:

    public Collection<String> getNames() {
        return new ArrayList<>() {{
            add(domain);
            if (subjectAltNames != null) {
                addAll(subjectAltNames);
            }
        }};
    }
```

The problem is at that point the Set seems to be initialized, so if no SAN is present the set is empty and not NULL

NiccoMlt:
- Added columns `attemptcount` and `message` to DB
- Added configuration `dynamiccertificatesmanager.errors.maxattempts` with default 10
- Added API at `/certificates/{domain}/reset` to reset certificate status to waiting, resetting also its counter and error message
- Added attempts counter and error message visualization in Certificate view:
  image

Closes #264, closes #359, closes #401


NiccoMlt:
Currently in draft to also try it manually a little better

carapace-server/src/main/java/org/carapaceproxy/core/RuntimeServerConfiguration.java
```diff
@@ -252,6 +253,9 @@ public void configure(ConfigurationStore properties) throws ConfigurationNotVali
         // storing enabled for all peers by default
         localCertificatesStorePeersIds = properties.getValues("dynamiccertificatesmanager.localcertificates.peers.ids");
         LOG.log(Level.INFO, "dynamiccertificatesmanager.localcertificates.peers.ids={0}", localCertificatesStorePeersIds);
+
+        maxAttempts = properties.getInt("carapage.maxattempts", maxAttempts);
```
hamadodene, line 257: This property is not dynamically updated if I change the maxattempts

carapace-server/src/main/java/org/carapaceproxy/server/certificates/DynamicCertificatesManager.java
```diff
@@ -337,26 +327,31 @@ private void certificatesLifecycle() {
...
-                            cert.setState(REQUEST_FAILED);
+                            cert.error("Order status for certificate is " + status);
                         }
                     }
-                    case REQUEST_FAILED -> { // challenge/order failed
+                    // challenge/order failed
+                    case REQUEST_FAILED -> {
                         LOG.log(Level.INFO, "Certificate issuing for domain: {0} current status is FAILED, setting status=WAITING again.", domain);
-                        cert.setState(WAITING);
+                        if (cert.getAttemptsCount() <= getConfig().getMaxAttempts()){
```
hamadodene, line 339: We can do the same thing when the certificate is always UNREACHBLE.

hamadodene (changes requested):

carapace-server/src/main/java/org/carapaceproxy/server/certificates/DynamicCertificatesManager.java
NiccoMlt: Makes sense

carapace-server/src/main/java/org/carapaceproxy/core/RuntimeServerConfiguration.java
NiccoMlt: Typo: _carapa_**g**_e.maxattempts_

NiccoMlt: Renamed to `dynamiccertificatesmanager.errors.maxattempts`

hamadodene (approved):

dmercuriali (approved):

NiccoMlt:
I wanted to add https://jsonassert.skyscreamer.org/ as a test dependency, but it is not available on the maven repositories specified in POM.
Maybe we could address it in an issue @hamadodene? Many tests rely on string comparisons on JSON to validate responses...

hamadodene:
> I wanted to add https://jsonassert.skyscreamer.org/ (https://github.com/skyscreamer/JSONassert) as a test dependency, but it is not available on the maven repositories specified in POM. Maybe we could address it in an issue @hamadodene? Many tests rely on string comparisons on JSON to validate responses...

Yes we can address it in a new issue. If you can, feel free to create the issue with all the details.

dmercuriali (approved):

Issue #415, opened 2023-02-27
Ability to use Epol EpollEventLoopGroup if supported by OS


hamadodene:
EpollEventLoopGroup vs. NioEventLoopGroupand in high load scenarios could provide better performance in:

- CPU usage
- Memory usage
- Allocations

It would be good to include the ability to switch to EpollEventLoopGroup if the operating system supports it.

PR #416, opened 2023-02-27
Use Epol EpollEventLoopGroup if supported by OS

hamadodene:
#415 

EpollEventLoopGroup vs. NioEventLoopGroupand in high load scenarios could provide better performance in:

- CPU usage
- Memory usage
- Allocations

It would be good to include the ability to switch to EpollEventLoopGroup if the operating system supports it.


hamadodene:
@pv3nturi I also made the keepAlive options configurable.
Added keepAlive options on server side as well.

pv3nturi (changes requested):
LGTM but you'd better to add a few tests to check for property names typo.

pv3nturi (approved):
+1

Issue #412, opened 2023-02-20
Carapace 1.8 > java.lang.NullPointerException


hamadodene:
In our production setup I have noticed several NPE.
I have no way to reproduce it.

```
eb 20, 2023 1:30:48 PM reactor.util.Loggers$Slf4JLogger error
SEVERE: [2e92224d-1, L:/xxxxxxxxxx:8089 - R:/xxxxxxxxxxx:43834]
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
        at java.base/java.util.regex.Matcher.getTextLength(Matcher.java:1769)
        at java.base/java.util.regex.Matcher.reset(Matcher.java:415)
        at java.base/java.util.regex.Matcher.<init>(Matcher.java:252)
        at java.base/java.util.regex.Pattern.matcher(Pattern.java:1134)
        at java.base/java.util.regex.Pattern.matches(Pattern.java:1175)
        at org.carapaceproxy.core.ProxyRequestsManager$ConnectionsManager.lambda$apply$4(ProxyRequestsManager.java:598)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
        at java.base/java.util.concurrent.ConcurrentHashMap$EntrySpliterator.tryAdvance(ConcurrentHashMap.java:3660)
        at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
        at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:647)
        at org.carapaceproxy.core.ProxyRequestsManager$ConnectionsManager.apply(ProxyRequestsManager.java:599)
        at org.carapaceproxy.core.ProxyRequestsManager.forward(ProxyRequestsManager.java:343)
[...]

PR #418, opened 2023-03-29
NullPointerException when http header host is null

hamadodene:
#412
As per RFC, if host header is missing or wrong the response must be 400 Bad request.
See https://www.rfc-editor.org/rfc/rfc2616#page-38
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

the error page can be customized by setting:
`default.action.badrequest=custom_action`

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequest.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequest.java
@@ -26,11 +26,14 @@ import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpScheme;
 import io.netty.handler.ssl.SslHandler;
+
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.concurrent.atomic.AtomicLong;
 import lombok.Data;
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.http.conn.util.InetAddressUtils;
 import org.carapaceproxy.server.config.NetworkListenerConfiguration.HostPort;
 import org.carapaceproxy.server.filters.UrlEncodedQueryString;
 import org.carapaceproxy.server.mapper.MapResult;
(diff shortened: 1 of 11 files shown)


dmercuriali (changes requested):
just a nit

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequest.java
```diff
@@ -171,6 +174,27 @@ public String getRequestHostname() {
...
+        int colonIndex = requestHostname.indexOf(":");
+        if (colonIndex != -1) {
+            String portString = requestHostname.substring(colonIndex + 1);
+            try {
+                int port = Integer.parseInt(portString);
+                if (port > 0 && port <= 65535) {
+                    return requestHostname.substring(0, colonIndex);
+                }
+            } catch (NumberFormatException e) {
+                return "";
```
pv3nturi, line 188: Maybe in that case is it better to return badRequest?

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequest.java
```diff
@@ -171,6 +174,27 @@ public String getRequestHostname() {
         return request.requestHeaders().get(HttpHeaderNames.HOST);
     }
 
+    public String getHostname() {
+        String requestHostname = getRequestHostname();
+        int colonIndex = requestHostname.indexOf(":");
```
pv3nturi, line 179: May this throw NPE?

pv3nturi (changes requested):

carapace-server/src/main/java/org/carapaceproxy/core/ProxyRequest.java
hamadodene: Empty hostname is already mapped to badRequest in `MapResult map(ProxyRequest request)` because it is an invalid hostname.

hamadodene:
@pv3nturi PTAL

pv3nturi (approved):

dmercuriali (approved):

PR #421, opened 2023-04-27
Fix direct memory leak

hamadodene:
In the current implementation when we want to save an object in cache we do a retainduplicated() of netty's bytebuf.
By saving the object in the cache it shares the count of direct memory with netty. However, this leads to a continuous allocation of direct memory by netty.

At some point, depending on the configured XX:MaxDirectMemorySize, netty will no longer be able to allocate new direct memory. Causing an OOM.

The Fix is to copy the Bytebuf into a ByteBuffer and release the Bytebuf of netty. This allows for separate counts between Cache and netty direct memory.

When you want to serve a response from the cache you wrap the ByteBuffer in a ByteBuf of netty so that it can be served to the final client.

To make clear that you license your contribution under 
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java
```diff
@@ -309,17 +311,12 @@ private void addChunk(ByteBuf chunk) {
         }
 
         void clear() {
-            chunks.forEach(ByteBuf::release);
```
pv3nturi, line 312: Are you sure we don't need to release each ByteBuffer?

pv3nturi (commented):
Just a comment

hamadodene:
@pv3nturi  Upgrade reactor netty for potential memory leak in netty 4.1.81.Final #424

pv3nturi (approved):
LGTM

dmercuriali (approved):

PR #437, opened 2023-05-16
Bump reactor-netty to 2022.0.7

pv3nturi:


hamadodene:
For test org.carapaceproxy.core.MaxHeaderSizeTest you have to change assertEquals(413, response2.statusCode()); to assertEquals(431, response2.statusCode());

See  https://github.com/reactor/reactor-netty/pull/2534

PR #446, opened 2023-10-30
Upgrade Node.js version to latest LTS

NiccoMlt:
Upgrade Node.js from Gallium LTS to Iron LTS.

The currently-used version is a problem with Mac on Apple M2 processors while building `node-gyp`.

Also, according to roadmap it is in end of life: <https://github.com/nodejs/Release>.


hamadodene:
@NiccoMlt Ok to me..
Can you fix tests?

@pv3nturi  PTAL

NiccoMlt:
> Can you fix tests?

Yes, I should, but test result collection is broken, I should fix it first...

hamadodene (approved):

Issue #456, opened 2024-02-14
stream has already been operated upon or closed when carapace try to choose a certifiacte for a request


hamadodene:
The issue occurs when dealing with two similar certificates, such as *.peachtest.it and *.qapatchweb.peachtest.it. When a request like test.qapatchweb.peachtest.it is made, and the initial match happens with *.qapatchweb.peachtest.it, the error is triggered. This is because the code attempts to reuse the same stream, which is not allowed.

```
Feb 14, 2024 11:51:44 AM org.carapaceproxy.server.certificates.DynamicCertificatesManager getCertificateForDomain
SEVERE: No dynamic certificate available for domain *
java.lang.IllegalStateException: stream has already been operated upon or closed
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:229)
        at java.base/java.util.stream.ReferencePipeline.anyMatch(ReferencePipeline.java:632)
        at org.carapaceproxy.server.config.SSLCertificateConfiguration.isMoreSpecific(SSLCertificateConfiguration.java:99)
        at org.carapaceproxy.core.Listeners.chooseCertificate(Listeners.java:455)
        at org.carapaceproxy.core.Listeners$ListeningChannel.lambda$map$0(Listeners.java:361)
        at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
        at org.carapaceproxy.core.Listeners$ListeningChannel.map(Listeners.java:359)
        at org.carapaceproxy.core.Listeners$ListeningChannel.map(Listeners.java:323)
        at io.netty.handler.ssl.SniHandler.lookup(SniHandler.java:97)
        at io.netty.handler.ssl.AbstractSniHandler.lookup(AbstractSniHandler.java:126)
        at io.netty.handler.ssl.SslClientHelloHandler.select(SslClientHelloHandler.java:188)
        at io.netty.handler.ssl.SslClientHelloHandler.decode(SslClientHelloHandler.java:149)
        at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:519)
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:458)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:280)
[...]

PR #457, opened 2024-02-14
Fix stream has already been operated upon or closed

hamadodene:
issue #456


carapace-server/src/main/java/org/carapaceproxy/server/config/SSLCertificateConfiguration.java
```diff
@@ -93,10 +93,13 @@ public boolean isMoreSpecific(SSLCertificateConfiguration other) {
...
-        final var otherNames = other.getNames().stream().map(CertificatesUtils::removeWildcard);
-        for (var n: getNames()) {
+
+        final var otherNamesList = other.getNames().stream().map(CertificatesUtils::removeWildcard)
+                .collect(Collectors.toList());
+
+        for (var n : getNames()) {
             final var name = CertificatesUtils.removeWildcard(n);
-            if (otherNames.anyMatch(on -> name.length() > on.length())) {
+            if (otherNamesList.stream().anyMatch(on -> name.length() > on.length())) {
```
NiccoMlt, line 102: Here couldn't we just store the max length of the name in otherNamesList and check against it, instead cycle through `otherNamesList` each time?

hamadodene: yes, make sense @NiccoMlt

NiccoMlt (approved):
Looks good to me

carapace-server/src/main/java/org/carapaceproxy/core/Listeners.java
```diff
@@ -148,6 +149,7 @@ private void stopListener(HostPort hostport) throws InterruptedException {
         ListeningChannel channel = listeningChannels.remove(hostport);
         if (channel != null) {
             channel.channel.disposeNow(Duration.ofSeconds(10));
+            FutureMono.from(channel.getConfig().getGroup().close()).block(Duration.ofSeconds(30));
```
NiccoMlt, line 152: I'm not sure if 30s is too much here.

Maybe we can go on with this, and change it or make it configurable if it becomes a problem

NiccoMlt (approved):
LGTM

carapace-server/src/main/java/org/carapaceproxy/core/Listeners.java
hamadodene: Maybe we can reduce to 10s?

NiccoMlt: It might be an improvement, for sure

NiccoMlt (approved):

PR #459, opened 2024-02-21
Fix PR validation permissions by splitting in two workflows

NiccoMlt:
Splitting PR validation workflow in two, one doing the actual validation, and the other publishing checks with the right permissions

See:

* https://github.com/marketplace/actions/junit-report-action#pr-run-permissions
* https://github.com/marketplace/actions/publish-test-results#permissions
* https://github.com/marketplace/actions/test-reporter#recommended-setup-for-public-repositories

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -16,10 +16,6 @@
 name: Java CI - Code tests
 on:
   pull_request:
-permissions:
-  contents: read
-  actions: read
-  checks: write
 env:
   MAVEN_OPTS: >-
     -Dmaven.artifact.threads=4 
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -48,11 +44,9 @@ jobs:
           cache: 'maven'
       - name: 'Build and test with Maven'
         run: mvn verify
-      - name: 'Report test results'
-        uses: dorny/test-reporter@v1
+      - uses: actions/upload-artifact@v3
```
NiccoMlt, line 47: We cannot use `actions/upload-artifact@v4` for now, see dorny/test-reporter#363

Issue #462, opened 2024-03-05
Support of Cache-Control header is only partial


NiccoMlt:
While looking at the block code below to figure out if WebP images would be cached correctly:

https://github.com/diennea/carapaceproxy/blob/fe1cfdfec3c52a8f004ad9676a3e3a386f71a131/carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java#L188-L200

we figured out that caching strategy implemented by `isChached` methods is a little naïve. Taking a look the the actual code:

https://github.com/diennea/carapaceproxy/blob/fe1cfdfec3c52a8f004ad9676a3e3a386f71a131/carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java#L122-L131

https://github.com/diennea/carapaceproxy/blob/fe1cfdfec3c52a8f004ad9676a3e3a386f71a131/carapace-server/src/main/java/org/carapaceproxy/server/cache/ContentsCache.java#L155-L172

it is especially evident that we don't actually support all the [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) values.

| Request          | Response                 |
|------------------|--------------------------|
| `max-age`        | `max-age`                |
| `max-stale`      | -                        |
| `min-fresh`      | -                        |
| -                | `s-maxage`               |
| `no-cache`       | `no-cache`               |
| `no-store`       | `no-store`               |
| `no-transform`   | `no-transform`           |
| `only-if-cached` | -                        |
| -                | `must-revalidate`        |
| -                | `proxy-revalidate`       |
| -                | `must-understand`        |
| -                | `private`                |
| -                | `public`                 |
| -                | `immutable`              |
| -                | `stale-while-revalidate` |
| `stale-if-error` | `stale-if-error`         |

PR #463, opened 2024-03-05
Add support to WebP caching

NiccoMlt:
Add support to WebP caching even in case of query string

See #462

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/configstore/ConfigurationStore.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/configstore/ConfigurationStore.java
@@ -21,10 +21,8 @@ package org.carapaceproxy.configstore;
 
 import herddb.utils.BooleanHolder;
 import java.security.KeyPair;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Properties;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


hamadodene:
@NiccoMlt  In my opinion, it makes sense to add a dynamic property to allow the user to potentially add other extensions as well.

NiccoMlt:
added `cache.cachealways` @hamadodene

NiccoMlt:
Fixed default as requested @hamadodene

hamadodene:
@NiccoMlt I don't know if what I think makes sense. But if I set cache.cachealways I also overwrite the defaults, right?
Maybe it's better if by setting the property they are added to the default? What do you think?

NiccoMlt:
Yes, it should
See https://github.com/diennea/carapaceproxy/blob/1b5d9864cad85f36dfdf1bfa276e3f6e375fafc2/carapace-server/src/main/java/org/carapaceproxy/core/RuntimeServerConfiguration.java#L280-L281

NiccoMlt:
Oh ok, I've read your comment only half-way.

I don't like the idea of not being able to disable the "always cached" behaviour if I want to.

I think it is better to configure the whole list.

hamadodene:
Make sense...LGTM

hamadodene (approved):

PR #466, opened 2024-03-28
Return Service Unavailable 503 when every backend is down

### LINKED ISSUE #436: Return 503 Service unavailable when all backend is unavailable
Currently when all backends are down we return 500 internal server error by default.
But I think we will have to return service unavailable.
500 internal error should be the errors in cases of problems on the Carapace side, such as:
- no action configured
- no backend configured
that is, errors internal to Carapace that are not attributable to problems on the backend side.

hamadodene:
Closes #436

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
@@ -19,34 +19,11 @@
  */
 package org.carapaceproxy.server.mapper;
 
-import static org.carapaceproxy.core.StaticContentsManager.DEFAULT_INTERNAL_SERVER_ERROR;
-import static org.carapaceproxy.core.StaticContentsManager.DEFAULT_MAINTENANCE_MODE_ERROR;
-import static org.carapaceproxy.core.StaticContentsManager.DEFAULT_NOT_FOUND;
-import static org.carapaceproxy.core.StaticContentsManager.IN_MEMORY_RESOURCE;
-import static org.carapaceproxy.server.config.DirectorConfiguration.ALL_BACKENDS;
-
 import io.netty.handler.codec.http.HttpResponseStatus;
-
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
 import org.carapaceproxy.SimpleHTTPResponse;
 import org.carapaceproxy.configstore.ConfigurationStore;
 import org.carapaceproxy.core.ProxyRequest;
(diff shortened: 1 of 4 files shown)


NiccoMlt (changes requested):
Looks good to me, but you need to fix a couple of tests:

- org.carapaceproxy.backends.StuckRequestsTest ► testBackendUnreachableOnStuckRequest(true)
- org.carapaceproxy.server.mapper.BasicStandardEndpointMapperTest ► testRouteErrors

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -276,13 +277,22 @@ public SimpleHTTPResponse mapInternalError(String routeid) {
...
         }
+        // custom global
+        if (errorAction == null && defaultServiceUnavailable != null) {
+            errorAction = actions.get(defaultServiceUnavailable);
+        }
+        if (errorAction != null) {
+            return new SimpleHTTPResponse(errorAction.getErrorCode(), errorAction.getFile(), errorAction.getCustomHeaders());
+        }
         // fallback
         return super.mapServiceUnavailableError(routeId);
```
NiccoMlt, line 297: We can leverage the `Optional` better here, i.e. like it follows:

```java
// custom for route
Optional<RouteConfiguration> config = routes.stream().filter(r -> r.getId().equalsIgnoreCase(routeid)).findFirst();
ActionConfiguration errorAction = config
        .map(RouteConfiguration::getErrorAction)
        .map(actions::get)
        // custom global
        .orElseGet(() -> defaultInternalErrorAction != null ? actions.get(defaultInternalErrorAction) : null);
if (errorAction != null) {
    return new SimpleHTTPResponse(errorAction.getErrorCode(), errorAction.getFile(), errorAction.getCustomHeaders());
}
// fallback
return super.mapInternalError(routeid);
```

carapace-server/src/main/java/org/carapaceproxy/server/mapper/StandardEndpointMapper.java
```diff
@@ -276,13 +277,22 @@ public SimpleHTTPResponse mapInternalError(String routeid) {
...
+                errorAction = actions.get(action);
             }
         }
+        // custom global
+        if (errorAction == null && defaultServiceUnavailable != null) {
+            errorAction = actions.get(defaultServiceUnavailable);
+        }
+        if (errorAction != null) {
+            return new SimpleHTTPResponse(errorAction.getErrorCode(), errorAction.getFile(), errorAction.getCustomHeaders());
+        }
```
NiccoMlt, line 295: This part is pretty much identical in:

- `mapInternalError`: lines 258-273
- `mapServiceUnavailableError`: lines 280-295

Maybe we should group it in an external method?

hamadodene:
@NiccoMlt test Fixed

NiccoMlt (changes requested):
Good job! Just a nitpick and we can go

NiccoMlt (approved):
Good job!

Issue #357, opened 2022-06-07
UI:Certificate> Implement a form for creating/modifying and deleting a certificate


hamadodene:

hamadodene:
Static/manual certificate creation, editing and deletion are missing.

PR #475, opened 2024-06-04
feat(server): add DELETE API for Carapace certificates

### LINKED ISSUE #454: Implement API Create/Delete and edit certificate
API for create/deleted and edit certificate

NiccoMlt:
Closes #454

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
@@ -48,6 +48,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.servlet.ServletContext;
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -55,6 +56,8 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import lombok.AllArgsConstructor;
@@ -67,6 +70,8 @@ import org.carapaceproxy.core.HttpProxyServer;
 import org.carapaceproxy.core.RuntimeServerConfiguration;
 import org.carapaceproxy.server.certificates.DynamicCertificateState;
 import org.carapaceproxy.server.certificates.DynamicCertificatesManager;
+import org.carapaceproxy.server.config.ConfigurationChangeInProgressException;
+import org.carapaceproxy.server.config.ConfigurationNotValidException;
 import org.carapaceproxy.server.config.SSLCertificateConfiguration;
 import org.carapaceproxy.server.config.SSLCertificateConfiguration.CertificateMode;
 import org.carapaceproxy.utils.CertificatesUtils;
@@ -80,11 +85,10 @@ import org.carapaceproxy.utils.CertificatesUtils;
 @Produces(MediaType.APPLICATION_JSON)
 public class CertificatesResource {
 
-    public static final Set<DynamicCertificateState> AVAILABLE_CERTIFICATES_STATES_FOR_UPLOAD = Set.of(AVAILABLE, WAITING);
     private static final Logger LOG = Logger.getLogger(CertificatesResource.class.getName());
 
-    @javax.ws.rs.core.Context
-    ServletContext context;
+    @Context
+    private ServletContext context;
 
     @Data
     @AllArgsConstructor
@@ -93,7 +97,7 @@ public class CertificatesResource {
         private final Collection<CertificateBean> certificates;
         private final String localStorePath;
 
-        public CertificatesResponse(Collection<CertificateBean> certificates, HttpProxyServer server) {
+        public CertificatesResponse(final Collection<CertificateBean> certificates, final HttpProxyServer server) {
             this.certificates = certificates;
             this.localStorePath = server.getCurrentConfiguration().getLocalCertificatesStorePath();
         }
@@ -165,7 +169,11 @@ public class CertificatesResource {
         return new CertificatesResponse(res.values(), server);
     }
 
-    private static void fillCertificateBean(CertificateBean bean, SSLCertificateConfiguration certificate, DynamicCertificatesManager dCManager, HttpProxyServer server) {
+    private static void fillCertificateBean(
+            final CertificateBean bean,
+            final SSLCertificateConfiguration certificate,
+            final DynamicCertificatesManager dCManager,
+            final HttpProxyServer server) {
         try {
             DynamicCertificateState state = null;
             if (certificate.isDynamic()) {
@@ -208,7 +216,7 @@ public class CertificatesResource {
 
     @GET
     @Path("{certId}")
-    public CertificatesResponse getCertificateById(@PathParam("certId") String certId) {
+    public CertificatesResponse getCertificateById(@PathParam("certId") final String certId) {
         final var cert = findCertificateById(certId);
         return new CertificatesResponse(
                 cert != null ? List.of(cert) : Collections.emptyList(),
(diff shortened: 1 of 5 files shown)


hamadodene (approved):
LGTM...
@NiccoMlt  Can you check the test that fails? It could be one of the flacky tests. If you can fix it it wouldn't be bad.

hamadodene:
Are you planning to integrate it on the admin side too?
Otherwise, can you make another follow up issue to integrate it if it isn't already there?

NiccoMlt:
> Are you planning to integrate it on the admin side too? Otherwise, can you make another follow up issue to integrate it if it isn't already there?

The issue already exists, the UI support was not planned for this issue: https://github.com/diennea/carapaceproxy/issues/357

> LGTM... @NiccoMlt Can you check the test that fails? It could be one of the flacky tests. If you can fix it it wouldn't be bad.

Ok, I'll give it a look

carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
```diff
@@ -261,25 +269,40 @@ public Response createCertificate(CertificateForm form) {
         return FormValidationResponse.created();
     }
 
+    @DELETE
+    @Path("{certId}")
+    public Response createCertificate(@PathParam("certId") final String certId) {
```
dmercuriali, line 274: should be `deleteCertificate`

carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
```diff
@@ -261,25 +269,40 @@ public Response createCertificate(CertificateForm form) {
...
     @Path("{certId}/download")
-    public Response downloadCertificateById(@PathParam("certId") String certId) throws GeneralSecurityException {
+    public Response downloadCertificateById(@PathParam("certId") final String certId) {
         CertificateBean cert = findCertificateById(certId);
-        byte[] data = new byte[0];
-        if (cert != null && cert.isDynamic()) {
-            HttpProxyServer server = (HttpProxyServer) context.getAttribute("server");
-            DynamicCertificatesManager dynamicCertificateManager = server.getDynamicCertificatesManager();
-            data = dynamicCertificateManager.getCertificateForDomain(cert.getId());
+        if (cert == null || !cert.isDynamic()) {
```
dmercuriali, line 293: This is a behaviour change. I'm ok with this but next time It would be best to create another PR.

We must remember to write this change in the release notes

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
```diff
@@ -475,6 +494,21 @@ public void saveCertificate(CertificateData cert) {
         }
     }
 
+    @Override
+    public void removeCertificate(final String certId) {
+        try {
+            final var connection = datasource.getConnection();
+            final var preparedStatement = connection.prepareStatement(REMOVE_DIGITAL_CERTIFICATES_TABLE);
```
dmercuriali, line 501: why not opening both in the twr? if the `prepareStatement` call fails the connection is not closed

dmercuriali (changes requested):

carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
NiccoMlt: Argh, you're right!

carapace-server/src/main/java/org/carapaceproxy/api/CertificatesResource.java
NiccoMlt: Actually, it only changed the error from a non-meaningful 500 originated from a NullPointerException to a more meaningful 404.

But you're right, I should've stated the updated somewhere

carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
NiccoMlt: The try-with-resources is in the line below: from Java 9 it should be an alternatively valid syntax

dmercuriali: It's valid syntax but I think the compiled code does not have the same meaning. By declaring the resources outside the twr, if the call to `connection#prepareStatement` fails then `connection#close` is not invoked.

PS. actually tested out of curiosity and I can confirm this. With this syntax the 2 resources are declared and then the 2 nested try are created. By declaring both inside the twr you end up with the first declaration and the first try block, which wraps the second declaration and the second try block

NiccoMlt: Cool, didn't know that, I totally missed the point of that addition, thank you!

I also tried it like that:

```java
public class Test {

    public interface Common extends AutoCloseable {
        @Override
        default void close() {
            System.out.println(this.getClass() + " closed correctly");
        }

        default Successful getSuccessful() {
            return new Successful();
        }
        
        default Failure getFailure() {
            return new Failure();
        }
    }

    public static class Successful implements Common {
        public Successful() {
            System.out.println(this.getClass() + " created correctly");
        }
    }

    public static class Failure implements Common {
        public Failure() {
            System.out.println(this.getClass() + " created");
            throw new IllegalStateException("Faiulre");
        }
    }

    public static void classic() {
        try (final var succ = new Successful(); final var fail = succ.getFailure()) {
            System.out.println("Try block");
        }
    }

    public static void java9() {
        final var succ = new Successful();
        try (final var fail = succ.getFailure()) {
            System.out.println("Try block");
        }
    }

    public static void likeIDid() {
        final var succ = new Successful();
        final var fail = succ.getFailure();
        try (succ; fail) {
            System.out.println("Try block");
        }
    }

    public static void main(final String[] args) {
        System.out.println("classic...");
        try {
            classic();
        } catch (Exception ignored) { }
        System.out.println("Java 9...");
        try {
            java9();
        } catch (Exception ignored) { }
        System.out.println("Like I did...");
        try {
            likeIDid();
        } catch (Exception ignored) { }
    }
}
```

which is compiled like that:

```java
//
// Source code recreated from a .class file by IntelliJ IDEA
// (powered by FernFlower decompiler)
//

public class Test {
    public Test() {
    }

    public static void classic() {
        Successful var0 = new Successful();

        try {
            Failure var1 = var0.getFailure();

            try {
                System.out.println("Try block");
            } catch (Throwable var6) {
                if (var1 != null) {
                    try {
                        var1.close();
                    } catch (Throwable var5) {
                        var6.addSuppressed(var5);
                    }
                }

                throw var6;
            }

            if (var1 != null) {
                var1.close();
            }
        } catch (Throwable var7) {
            try {
                var0.close();
            } catch (Throwable var4) {
                var7.addSuppressed(var4);
            }

            throw var7;
        }

        var0.close();
    }

    public static void java9() {
        Successful var0 = new Successful();
        Failure var1 = var0.getFailure();

        try {
            System.out.println("Try block");
        } catch (Throwable var5) {
            if (var1 != null) {
                try {
                    var1.close();
                } catch (Throwable var4) {
                    var5.addSuppressed(var4);
                }
            }

            throw var5;
        }

        if (var1 != null) {
            var1.close();
        }

    }

    public static void likeIDid() {
        Successful var0 = new Successful();
        Failure var1 = var0.getFailure();
        Successful var2 = var0;

        try {
            Failure var3 = var1;

            try {
                System.out.println("Try block");
            } catch (Throwable var8) {
                if (var1 != null) {
                    try {
                        var3.close();
                    } catch (Throwable var7) {
                        var8.addSuppressed(var7);
                    }
                }

                throw var8;
            }

            if (var1 != null) {
                var1.close();
            }
        } catch (Throwable var9) {
            if (var0 != null) {
                try {
                    var2.close();
                } catch (Throwable var6) {
                    var9.addSuppressed(var6);
                }
            }

            throw var9;
        }

        if (var0 != null) {
            var0.close();
        }

    }

    public static void main(String[] var0) {
        System.out.println("classic...");

        try {
            classic();
        } catch (Exception var4) {
        }

        System.out.println("Java 9...");

        try {
            java9();
        } catch (Exception var3) {
        }

        System.out.println("Like I did...");

        try {
            likeIDid();
        } catch (Exception var2) {
        }

    }

    public static class Successful implements Common {
        public Successful() {
            System.out.println(String.valueOf(this.getClass()) + " created correctly");
        }
    }

    public static class Failure implements Common {
        public Failure() {
            System.out.println(String.valueOf(this.getClass()) + " created");
            throw new IllegalStateException("Faiulre");
        }
    }

    public interface Common extends AutoCloseable {
        default void close() {
            System.out.println(String.valueOf(this.getClass()) + " closed correctly");
        }

        default Successful getSuccessful() {
            return new Successful();
        }

        default Failure getFailure() {
            return new Failure();
        }
    }
}
```

and in fact, if run, gives the behavior that you were hinting:

```
❯ java Test                                                                                                          
classic...
class Test$Successful created correctly
class Test$Failure created
class Test$Successful closed correctly
Java 9...
class Test$Successful created correctly
class Test$Failure created
Like I did...
class Test$Successful created correctly
class Test$Failure created
```

Thank you for pointing it out, I'll fix it

NiccoMlt:
> > LGTM... @NiccoMlt Can you check the test that fails? It could be one of the flacky tests. If you can fix it it wouldn't be bad.
> 
> Ok, I'll give it a look

@hamadodene sorry but I didn't manage to reproduce it in any way, it's easier to break it for "too many files" than to have the same problem happen locally:

image

PR #477, opened 2024-06-06
feat(server): Trust `X-Forwarded-For` header for some IPs

NiccoMlt:
- Add [Apache `commons-net:commons-net:3.11.0`](https://commons.apache.org/proper/commons-net/) dependency
- Move [`X-Forwarded-For` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) management from `add-x-forwarded-for` filter to new listener options:
  - `add-x-forwarded-for` filter is now deprecated for removal
  - added `forwarded` = `DROP` | `PRESERVE` | `REWRITE` (default) | `IF_TRUSTED` listener option
  - added `trustedips` optional listener option, that accepts CIDR that should be considered trusted

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/server/config/NetworkListenerConfiguration.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/server/config/NetworkListenerConfiguration.java
@@ -19,83 +19,85 @@
  */
 package org.carapaceproxy.server.config;
 
-import java.util.Collections;
-import java.util.Set;
-
 import io.netty.channel.group.ChannelGroup;
 import io.netty.channel.group.DefaultChannelGroup;
 import io.netty.util.concurrent.DefaultEventExecutor;
-import lombok.Data;
+import java.util.Set;
+import lombok.Getter;
+import org.carapaceproxy.core.ForwardedStrategies;
 
 /**
  * Listens for connections on the network
  */
-@Data
+@Getter
 public class NetworkListenerConfiguration {
 
     public static final Set<String> DEFAULT_SSL_PROTOCOLS = Set.of("TLSv1.2", "TLSv1.3");
+    public static final int DEFAULT_SO_BACKLOG = 128;
+    public static final int DEFAULT_KEEP_ALIVE_IDLE = 300;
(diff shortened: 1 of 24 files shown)


dmercuriali (approved):

carapace-server/src/main/java/org/carapaceproxy/server/config/NetworkListenerConfiguration.java
```diff
@@ -19,83 +19,85 @@
...
+@Getter
 public class NetworkListenerConfiguration {
 
     public static final Set<String> DEFAULT_SSL_PROTOCOLS = Set.of("TLSv1.2", "TLSv1.3");
+    public static final int DEFAULT_SO_BACKLOG = 128;
+    public static final int DEFAULT_KEEP_ALIVE_IDLE = 300;
+    public static final int DEFAULT_KEEP_ALIVE_INTERVAL = 60;
+    public static final int DEFAULT_KEEP_ALIVE_COUNT = 8;
+    public static final int DEFAULT_MAX_KEEP_ALIVE_REQUESTS = 1000;
+    public static final String DEFAULT_FORWARDED_STRATEGY = ForwardedStrategies.preserve().name();
```
hamadodene, line 41: In the PR description, you wrote that the default is REWRITE, but here you use PRESERVE

hamadodene (changes requested):

carapace-server/src/main/java/org/carapaceproxy/server/config/NetworkListenerConfiguration.java
hamadodene: It is preferable to have REWRITE as the default.

NiccoMlt: > In the PR description, you wrote that the default is REWRITE, but here you use PRESERVE

Oh, you're right, sorry!

> It is preferable to have REWRITE as the default.

OK

hamadodene: Make sense..ok to set default to preserve

Issue #409, opened 2023-02-17
HTTP/2 - Enable http2 H2C 


hamadodene:
As the first step of  #181   , we can implement http2 H2C support.
See https://projectreactor.io/docs/netty/release/reference/index.html#_http2 for more info.

It must be possible on the listener side to specify which http protocol to use. Something like:
```
listener.1.protocol=http11,h2c
or
listener.1.protocol=h2c
or
listener.1.protocol=http11
```

H2C will need to be supported for http listeners only

hamadodene:
Original issue #181

hamadodene:
Note: Http2 will also need to be supported on the client side.

hamadodene:
https://github.com/reactor/reactor-netty/pull/2659
reactor netty 1.1.3 introduce possibility to configure HTT11, H2 e H2C together

PR #490, opened 2024-08-30
HTTP/2 - Enable H2 protocol

### LINKED ISSUE #181: HTTP/2 - Enable HTTP/2, ALPN & NPN by default
For performance reason enable by default HTTP/2, Application Layer Protocol Negotiation and Next Protocol Negotiation.

We should have all the building blocks (Netty, boringSSL) to do so

### LINKED ISSUE #410: HTTP/2 - Enable http2 H2
As the second step of https://github.com/diennea/carapaceproxy/issues/181 , we can implement http2 H2 support.
See https://projectreactor.io/docs/netty/release/reference/index.html#_http2 for more info.

As with #409, it must be possible to specify H2 protocol support on the listener.
H2 will need to be supported for https listeners only.

### LINKED ISSUE #506: Listeners > Listener not booting when certificate is unreachable
While attempting to update Carapace with the branch 410-http2-enable-http2-h2, Carapace fails to start due to this error:
```
SEVERE: No dynamic certificate available for domain cara8testxx.example.it
Oct 28, 2024 9:20:55 AM org.carapaceproxy.core.ListeningChannel bootSslContext
SEVERE: ERROR booting listener
java.io.IOException: keystore password was incorrect
        at java.base/sun.security.pkcs12.PKCS12KeyStore.engineLoad(PKCS12KeyStore.java:2097)
        at java.base/sun.security.util.KeyStoreDelegator.engineLoad(KeyStoreDelegator.java:228)
        at java.base/java.security.KeyStore.load(KeyStore.java:1500)
        at org.carapaceproxy.utils.CertificatesUtils.loadKeyStoreData(CertificatesUtils.java:179)
        at org.carapaceproxy.core.ListeningChannel.bootSslContext(ListeningChannel.java:140)
        at org.carapaceproxy.core.ListeningChannel.map(ListeningChannel.java:106)
        at org.carapaceproxy.core.ListeningChannel.applySslContext(ListeningChannel.java:213)
        at org.carapaceproxy.core.Listeners.lambda$bootListener$1(Listeners.java:199)
        at reactor.netty.http.server.HttpServer.secure(HttpServer.java:807)
        at reactor.netty.http.server.HttpServer.secure(HttpServer.java:776)
        at org.carapaceproxy.core.Listeners.bootListener(Listeners.java:191)
        at org.carapaceproxy.core.Listeners.reloadConfiguration(Listeners.java:165)
        at org.carapaceproxy.core.Listeners.start(Listeners.java:101)
        at org.carapaceproxy.core.HttpProxyServer.start(HttpProxyServer.java:446)
        at org.carapaceproxy.launcher.ServerMain.start(ServerMain.java:181)
        at org.carapaceproxy.launcher.ServerMain.main(ServerMain.java:129)
Caused by: java.security.UnrecoverableKeyException: failed to decrypt safe contents entry: javax.crypto.BadPaddingException: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
        ... 16 more
```

The certificate in question, which is being loaded, is actually in an UNREACHABLE state, meaning there is no certificate, or there may not be a certificate available for this domain.

Therefore, we need to ensure that we load ONLY certificates that are in the AVAILABLE state.

NiccoMlt:
Closes #410 and, along with #409, closes #181
Closes #506


hamadodene:
@NiccoMlt Take a look on this issue https://github.com/diennea/carapaceproxy/issues/506
I also get this error:
```
Oct 28, 2024 9:20:55 AM org.carapaceproxy.core.ListeningChannel map
SEVERE: Error booting certificate for SNI hostname cara17test.xxx.it, on listener NetworkListenerConfiguration[host=0.0.0.0, port=4089, ssl=true, sslCiphers=, defaultCertificate=*, sslProtocols=[TLSv1.3], soBacklog=128, keepAlive=true, keepAliveIdle=300, keepAliveInterval=60, keepAliveCount=8, maxKeepAliveRequests=10, forwardedStrategy=IF_TRUSTED, trustedIps=[127.0.0.1], protocols=[H2], group=DefaultChannelGroup(name: group-0x2, size: 0)]
```
The certificato of domain cara17test.xxx.it is available. I don't have the full stack. I will try to debug it later

NiccoMlt:
Sure

carapace-server/src/main/java/org/carapaceproxy/core/Listeners.java
NiccoMlt, line 1: `Listeners` was pretty much rewritten from the ground up a couple of times in the later months, but I tried to strip most modifications and keep it simple.

`ListeningChannel` nested class made it hard to track the flow of the setup of the channel and was externalized. More aggressive refactors were tried but dropped. We still need to find a way to wrap the "concept of channel" inside of it, while keeping the common setup into `Listeners`.

NiccoMlt: The behaviour wasn't changed much here: the SSL context setup was moved from `reactor.netty.http.server.HttpServer#doOnChannelInit` to `reactor.netty.http.server.HttpServer#secure(java.util.function.Consumer<? super reactor.netty.tcp.SslProvider.SslContextSpec>)`.

This poses one major change: `SslContext`s are loaded once, and then the listener needs to be restarted to pick up changes into the certificate. Hooks are already in place to take care of that, we only added a check on changes on the certs in addition to listener confs.

carapace-server/src/main/java/org/carapaceproxy/core/ListeningChannel.java
NiccoMlt, line 1: This class is not actually new but moved out of `Listeners` to track clearly the dependencies.
This class handles SSL contexts build, taking care of OCSP stapling (which is now done through a `handlerConfigurator`.

This class keeps some problems, that were not evident when it was a nested, non-static, class.
For example, different listeners will load pretty much the same certificates but with their own SSL ciphers and protocols. This might lead to tricky behaviors.
This was mitigated by not actually sharing the certs, but leads to duplication, and I think that these settings should simply be moved from listerners to certs.

carapace-server/src/main/java/org/carapaceproxy/core/OcspSslHandler.java
NiccoMlt, line 1: This is pretty much a glorified lambda that actually does the stapling.
The original beahviour (which I note that comes from the [netty example](https://github.com/netty/netty/tree/4.1/example/src/main/java/io/netty/example/ocsp) and adjusted for ReactorNetty) was not changed in this PR, I think.

carapace-server/src/main/java/org/carapaceproxy/server/config/NetworkListenerConfiguration.java
```diff
@@ -107,64 +113,15 @@ public NetworkListenerConfiguration(
...
-                .map(String::toUpperCase)
-                .map(HttpProtocol::valueOf)
-                .collect(Collectors.toUnmodifiableSet());
-        if (!ssl && httpProtocols.contains(H2)) {
-            throw new IllegalArgumentException("H2 requires SSL support");
-        }
-        this.protocols = httpProtocols;
+    public static Set<HttpProtocol> getDefaultHttpProtocols(final boolean ssl) {
+        return Set.of(HTTP11, ssl ? H2 : H2C);
     }
```
NiccoMlt, line 122: Now we enable HTTP/2 by default with SSL, or HTTP/2 with cleartext if without

carapace-server/src/test/java/org/carapaceproxy/ApplyConfigurationTest.java
NiccoMlt, line 1: This was rewritten to debug a problem in the past, then merged to a refactor that was introduced with one of the previous PR.
Refactors were done and were kept.

NiccoMlt (commented):
Left a couple of comments to help review @hamadodene @dmercuriali

dmercuriali (approved):

PR #473, opened 2024-05-06
Implement CRUD APIs for connection pools

NiccoMlt:
See #419 

### Maven wrapper

During the CI pipeline, I experienced some problems between Java 21 and the version of the `maven-compiler-plugin` bundled with the version of maven provided with the GitHub Actions.

To pin a specific version of Maven, I decided to introduce the [wrapper](https://maven.apache.org/wrapper): like Gradle, it introduces a couple of shell scripts (plus a Java one) that handle the download and the execution of a specific version of the build system, in fact making it another dependency of the project.

This is good to control better how the build is done, both in CI and in development.

### GitHub Actions workflow

- Java version was updated to Java 21
- Now using Maven wrapper instead of Maven directly
- Updated dorny/test-reporter action to version v1.9.1, that allows to use v4 of actions/upload-artifact and actions/checkout, which were deprecated

### JUnit 5 Vintage dependency excluded

see https://github.com/diennea/carapaceproxy/issues/440

PR #494, opened 2024-09-10
fix(server): Use default value of Connection Pool when property is no…

### LINKED ISSUE #491: Connection Pool > Use default value when property is not set
When I create a connection pool via API, passing the following parameters in the JSON, for example:
```
{
    "id": "test-pool",
    "domain": ".*test.*",
    "maxConnectionsPerEndpoint": 10,
    "borrowTimeout": 60000,
    "connectTimeout": 10000,
    "stuckRequestTimeout": 120000,
    "idleTimeout": 60000,
    "maxLifeTime": 100000,
    "enabled": true
}
```
When the pool is created, the result is as follows:

```
connectionpool.3.borrowtimeout=60000
connectionpool.3.connecttimeout=10000
connectionpool.3.disposetimeout=0
connectionpool.3.domain=.*test.*
connectionpool.3.enabled=true
connectionpool.3.id=test-pool
connectionpool.3.idletimeout=60000
connectionpool.3.keepalive=false
connectionpool.3.keepalivecount=0
connectionpool.3.keepaliveidle=0
connectionpool.3.keepaliveinterval=0
connectionpool.3.maxconnectionsperendpoint=10
connectionpool.3.maxlifetime=100000
connectionpool.3.stuckrequesttimeout=120000
```

We can observe that some properties I didn’t specify, such as keepalivecount, are set to 0.
For these properties, we already have default values defined in `RuntimeServerConfiguration.java`. It would be appropriate to use the default values if the user doesn’t specify anything in the JSON.

NiccoMlt:
Follow-up of #473, closes #491

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/api/ConnectionPoolsResource.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/api/ConnectionPoolsResource.java
@@ -21,6 +21,17 @@ package org.carapaceproxy.api;
 
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.summingInt;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_BORROW_TIMEOUT;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_CONNECT_TIMEOUT;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_DISPOSE_TIMEOUT;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_IDLE_TIMEOUT;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_KEEPALIVE;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_KEEPALIVE_COUNT;
+import static org.carapaceproxy.server.config.ConnectionPoolConfiguration.DEFAULT_KEEPALIVE_IDLE;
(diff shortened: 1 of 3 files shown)


carapace-server/src/main/java/org/carapaceproxy/api/ConnectionPoolsResource.java
```diff
@@ -67,16 +77,16 @@ public static final class ConnectionPoolBean {
...
+        private int borrowTimeout = DEFAULT_BORROW_TIMEOUT;
+        private int connectTimeout = DEFAULT_CONNECT_TIMEOUT;
+        private int stuckRequestTimeout = DEFAULT_STUCK_REQUEST_TIMEOUT;
+        private int idleTimeout = DEFAULT_IDLE_TIMEOUT;
+        private int maxLifeTime = DEFAULT_MAX_LIFETIME;
+        private int disposeTimeout = DEFAULT_DISPOSE_TIMEOUT;
+        private int keepaliveIdle = DEFAULT_KEEPALIVE_IDLE;
+        private int keepaliveInterval = DEFAULT_KEEPALIVE_INTERVAL;
+        private int keepaliveCount = DEFAULT_KEEPALIVE_COUNT;
         private boolean keepAlive;
```
hamadodene, line 90: @NiccoMlt add also keepAlive=true as default

hamadodene (changes requested):

carapace-server/src/main/java/org/carapaceproxy/api/ConnectionPoolsResource.java
NiccoMlt: done

hamadodene (approved):

PR #512, opened 2024-11-07
Insert a rate limit for cold backends

NiccoMlt:
# Desired 

We need to model the spectrum between down backends and up backends.
We need some way to describe a backend that is "cold", meaning it was down and it got up since less than some, configurable, amount of time.

Add a new configuration should be added to define how much time should pass for a cold backend to warm up; for simplicity I will add it globally, and not per backend.

Also, a new configuration should be added to define how many connections a cold backend is able to handle; this should be a per-backend configuration instead...

# Implementation

- Added configuration `healthmanager.warmupperiod` (default **`30`s**) → time needed to go from `COLD` to `STABLE`
- Added configuration `healthmanager.tolerant` (default **`false`**) → if true, admits connections even beyond `safeCapacity` if needed
- Added configuration backends.<id>.safeCapacity (default **`10` connections**) → number of connections allowed by the backend when `COLD`


carapace-server/src/main/java/org/carapaceproxy/server/backends/BackendHealthStatus.java
```diff
@@ -19,75 +19,122 @@
...
+        /**
+         * The backend is reachable, but not since long.
+         * When in this safe, it is reasonable to assume that it is still warming-up,
+         * so it would be a sensible decision not to overload it with requests.
+         */
+        COLD,
+        /**
+         * The backend is reachable and was so since a reasonably-long time.
+         */
+        WARM
```
hamadodene, line 138: If the backend is relatively stable for a sufficiently long time, perhaps we could use something like:
ACTIVE, HEALTHY, or STABLE.
In this case, STABLE might sound better as the focus is on stability over time.

What do you think @NiccoMlt  ?

NiccoMlt: I agree, I was out of ideas on which name to use here ahah

NiccoMlt: So we want to call them:

- `DOWN`
- `COLD`
- `STABLE`

Right?

hamadodene: Yes, can go for me

hamadodene (approved):
LGTM... @dmercuriali  can you also take a look?

NiccoMlt:
Thank you for the review @hamadodene.
I'm still reworking a bunch of things, like moving out the JSpecify introduction on a dedicated PR and improving the backend selection logic.
I think i will work on it until wednesday, then we can integrate whatever is up on this branch

carapace-server/src/main/java/org/carapaceproxy/client/EndpointConnection.java
```diff
@@ -22,6 +22,7 @@
 import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.LastHttpContent;
+import org.carapaceproxy.core.EndpointKey;
```
dmercuriali, line 25: is this necessary?

carapace-server/src/main/java/org/carapaceproxy/client/ConnectionsManagerStats.java
```diff
@@ -21,6 +21,7 @@
 
 import java.util.Map;
 import org.carapaceproxy.EndpointStats;
+import org.carapaceproxy.core.EndpointKey;
```
dmercuriali, line 24: is this necessary?

carapace-server/src/main/java/org/carapaceproxy/client/ConnectionsManager.java
```diff
@@ -19,6 +19,7 @@
  */
 package org.carapaceproxy.client;
 
+import org.carapaceproxy.core.EndpointKey;
```
dmercuriali, line 22: is this necessary?

dmercuriali (changes requested):
Most of this PR is refactoring and restyling. 

We should split this in 2 PR, one with refactors and one with the actual implementation.

NiccoMlt:
I do not agree that **most** of it is refactor, but yes, it was maybe more than strictly necessary.

I tried to stay neat with the commits, I can try to split up the PR

NiccoMlt:
Merge after #516

dmercuriali (changes requested):
needs a rebase

dmercuriali (approved):

PR #513, opened 2024-11-08
build: migrate log backend to slf4j

NiccoMlt:
- migrate `java.util.logging` usages to SLF4J
- drop Apache Commons Logging dependency
- add Apache Commons Logging to SLF4J adapter because of other dependencies
- drop JUL backend for SLF4J
- add Logback as backend for SLF4J


hamadodene (approved):
Thanks @NiccoMlt ... Look good to me...
Please add also a note in the README or in another place for migration from logger to SL4J when user will install this PR.

hamadodene:
> I don't think that documenting it in the README would be useful, as the logger is an implementation detail that we already are **not** talking about.
> 
> I think we should document it in the changelog when we release it.
> 
> What do you think?

Make sense...thanks

carapace-server/src/main/java/org/carapaceproxy/utils/CarapaceLogger.java
```diff
@@ -41,7 +42,15 @@ public static void setLoggingDebugEnabled(boolean loggingDebugEnabled) {
...
+    /**
+     * This method assumes that {@link Level#DEBUG debug log level} is disabled.
+     * If {@link #setLoggingDebugEnabled(boolean) Carapace log level was enabled},
+     * forwards the log event to the {@link Level#INFO info log level} instead of using the debug level.
+     *
+     * @param s the format string
+     * @param o a list of arguments
+     */
+    public static void debug(String s, Object... o) {
+        (loggingDebugEnabled ? LOG.atInfo() : LOG.atDebug()).log(s, o);
```
NiccoMlt, line 54: I don't like this tricky approach at all, but I couldn't find any better approach that wouldn't break current `logging.debug.enabled` configuration, so I kept this approach.

If we ever think about revamping the configuration, I think we should drop this class and just leverage logback configuration (or the native configuration of whatever backend we will choose to put in our classpath)

NiccoMlt: Take a look at this link to the Logback docs: <https://logback.qos.ch/manual/configuration.html#autoScan>

Is "Automatically reloading configuration file upon modification" what you are looking for?

hamadodene: @NiccoMlt Yes this is what i mean...it's already implemented?

NiccoMlt: Yes, kinda; this is a built-in Logback feature.

Logback's auto-scan feature automatically detects changes to the logging configuration file and reloads the configuration without restarting the application. If the configuration file (e.g., `logback.xml`) is modified, Logback automatically reloads it. If the configuration file contains errors during the reload, Logback keeps the current configuration.

The point is, to enable it we need to include the `scan` attribute in the `<configuration>` element, and we are currently **not** doing so, but I certainly enable it; then, when we update `logback.xml` **in the classpath** we get it auto-reloaded.

This obviously means that if we launch a self-contained Carapace JAR, we cannot edit it (because the resource is inside the archive). A possible workaround would be to use an external configuration file with the `logback.configurationFile` property.

NiccoMlt: Done!

- dropped the `logging.debug.enabled` property, replacing the logs with debugs
- enabled `scan` on the default configuration

NiccoMlt:
To reduce the chance of merge conflicts, consider integrating this after #515 and before #516

dmercuriali (approved):

dmercuriali:
@NiccoMlt +1 but a rebase is needed

NiccoMlt:
@dmercuriali thanks, I'll do it later this afternoon

PR #522, opened 2025-01-18
ci: wrong hotfix version computed

### LINKED ISSUE #498: CI: wrong hotfix version computed
# Problem

Release job for hotfix 2.0.1 acutally computed a version 2.1.0: 
https://github.com/diennea/carapaceproxy/actions/runs/10795758991

```bash
./mvnw release:clean release:prepare release:perform \
      -Dtag=v2.1.0 \
      -DreleaseVersion=2.1.0 \
      -DdevelopmentVersion=2.1.1-SNAPSHOT \
      -Darguments='-DskipTests=true'
```

# Why it happened

The POM file on `release/2.0` was on version `2.1.0-SNAPSHOT`

https://github.com/diennea/carapaceproxy/blob/57f4482a3f9fbe1cb818159340600d3ad87146a8/pom.xml#L7

In the 'Compute versions and branch name' step, for hotfixes:

https://github.com/diennea/carapaceproxy/blob/57f4482a3f9fbe1cb818159340600d3ad87146a8/.github/workflows/release.yml#L51-L56

the code extracts the major, minor, and patch versions from the current POM version. For a hotfix, it uses these values directly for the release version.

The problem is that:
   - The POM version is `2.1.0-SNAPSHOT`
   - The script extracts major=2, minor=1, patch=0
   - It then sets the release_version to "2.1.0"
   - The hotfix process assumes that the version in the release branch's POM is already incremented to the next minor version (2.1.0-SNAPSHOT in this case)
   - It doesn't decrement the minor version or check if it's on a release branch with a different version number

For a proper hotfix, I think that:
   - The release/2.0 branch should have had a POM version of `2.0.1-SNAPSHOT`
   - Then the hotfix release would correctly become 2.0.1

NiccoMlt:
fixes #498

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ on:
           - major
           - minor
           - hotfix
-        default: major
+        default: minor
 
 permissions:
   contents: write
@@ -26,19 +26,26 @@ jobs:
     steps:
       - name: 'Checkout ${{ github.ref_name }} branch'
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref_name }}
+
       - name: 'Set up JDK 21'
         uses: actions/setup-java@v4
         with:
           java-version: '21'
           distribution: 'temurin'
           cache: 'maven'
+
       - name: 'Configure Git'
         run: |
           git config user.name "Diennea[bot]"
           git config user.email "172403682+diennea-bot@users.noreply.github.com"
-      - uses: webfactory/ssh-agent@v0.9.0
+
+      - name: 'Inject machine user SSH key into SSH agent'
+        uses: webfactory/ssh-agent@v0.9.0
         with:
           ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+
       - id: compute_versions
         name: 'Compute versions and branch name'
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -47,60 +54,69 @@ jobs:
...
             release_version="${new_major}.0.0"
             release_development_version="${new_major}.0.1-${qualifier}"
+            master_development_version="${new_major}.1.0-${qualifier}"
             ;;
           minor)
             new_minor=$((minor + 1))
-            branch_name="release/${major}.${new_minor}"
-            release_version="${major}.${new_minor}.0"
+            branch_name="release/${major}.${minor}"
+            release_version="${major}.${minor}.0"
```
NiccoMlt, line 75: Release version is `${major}.${minor}.${patch}` because we should expect `master` branch already be on version `x.(y+1).0-SNAPSHOT` when latest released version is `x.y.z`

.github/workflows/release.yml
```diff
@@ -47,60 +54,69 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          ./mvnw release:clean release:prepare release:perform \
-              -Dtag=${{ steps.compute_versions.outputs.tag }} \
+          ./mvnw release:clean release:prepare \
               -DreleaseVersion=${{ steps.compute_versions.outputs.release_version }} \
               -DdevelopmentVersion=${{ steps.compute_versions.outputs.release_development_version }} \
-              -Darguments='-DskipTests=true'
+              -Dtag=${{ steps.compute_versions.outputs.tag }}
```
NiccoMlt, line 112: The actual release preparation process should be shared between major, minor, and hotfix releases, because it always happens onto the release branch.

The version computation is what actually changes (besides the need to start a release branch for major and minor releases); I thought about tracking it into Maven profiles, but it was not easily feasible (if at all).

pom.xml
```diff
@@ -75,7 +75,7 @@
         <url>git@github.com:diennea/carapaceproxy.git</url>
         <connection>scm:git:${project.scm.url}</connection>
         <developerConnection>scm:git:${project.scm.url}</developerConnection>
-        <tag>release/1.11</tag>
+        <tag>HEAD</tag>
```
NiccoMlt, line 78: `tag` value should be managed by the `maven-release-plugin` goals, and describe what actually is the commit status:

- `HEAD` on master
- `release/<x>.<y>` on release branches
- `v<x>.<y>.<z>` on tagged commit of release branch, that are the actually released points in history

hamadodene (approved):

PR #593, opened 2026-06-05
fix: guard empty URI in server metrics to avoid crash on malformed requests

NiccoMlt:
## Context

When a malformed/undecodable HTTP request arrives (typically scanner or bot traffic over TLS), reactor-netty's `HttpTrafficHandler` writes a `400` and, with server metrics enabled, `MicrometerHttpServerMetricsHandler` runs on that response. For a decode failure `resolvePath(...)` returns an empty string, and the handler then calls `path.substring(1)`, throwing on the event loop:

```
java.lang.StringIndexOutOfBoundsException: Range [1, 0) out of bounds for length 0
    at java.base/java.lang.String.substring(String.java:2807)
    at reactor.netty.http.server.MicrometerHttpServerMetricsHandler$ResponseTimeHandlerContext.<init>(...)
    at reactor.netty.http.server.HttpServerOperations.sendDecodingFailures(...)
    at org.carapaceproxy.core.Listeners$2.channelRead(Listeners.java)
```

The connection is dropped, the metric is lost, and the error recurs on every such request. The URI tag was also configured with `Function.identity()`, which makes every distinct raw URI its own metric tag — a cardinality hazard. Observed firing in a live deployment against scanner/bot traffic on the HTTPS listener.

## Approach

The path flows through the `uriTagValue` function before reaching `substring(1)`, so this is fixable on the Carapace side without touching reactor-netty — map empty/`null` URIs to `"/"`:

```java
.metrics(true, uri -> uri == null || uri.isEmpty() ? "/" : uri)
```

This also templates the tag, removing the cardinality issue. reactor-netty 1.2.x still carries the unguarded `substring(1)`, so upgrading within that line does not address it.

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/core/Listeners.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/core/Listeners.java
@@ -36,7 +36,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.function.Function;
 import jdk.net.ExtendedSocketOptions;
 import org.carapaceproxy.server.config.ConfigurationNotValidException;
 import org.carapaceproxy.server.config.NetworkListenerConfiguration;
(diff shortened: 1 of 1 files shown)


NiccoMlt:
@coderabbitai resume

dmercuriali (approved):

PR #595, opened 2026-06-05
fix: correct backend-health state and per-request connection/metric accounting

NiccoMlt:
## Context

The per-request bookkeeping that feeds backend selection and metrics has several correctness bugs that compound under load: the pending-request gauge and the per-backend connection counter can double-count or drift negative, the `BackendHealthStatus` state machine mutates several fields non-atomically, and `SafeBackendSelector`'s cold-backend capacity check is inverted. Access logging for proxied/cached requests also fires at assembly time rather than on the reactive terminal, so the logged status and timing are wrong.

## Approach

- **Inverted safe-capacity check** (`BackendHealthManager`): `exceedsCapacity` compared `safeCapacity > connections`, deprioritising COLD backends that are within their limit and preferring overloaded ones — the opposite of warmup rate-limiting. Flipped to `connections > safeCapacity`.
- **Counter double-decrement** (`ProxyRequestsManager`): `PENDING_REQUESTS_GAUGE` and `healthStatus.connections` were decremented in both the success and the error callback; when response headers succeed but the body stream errors, both fire and the counters drift negative. Decrement exactly once, on the returned publisher's `doFinally`.
- **TOCTOU on decrement** (`BackendHealthStatus`): replaced check-then-`decrementAndGet` with an atomic `updateAndGet(v -> Math.max(0, v - 1))`.
- **Non-atomic health transitions** (`BackendHealthStatus`): the four state fields (`status`, `unreachableSince`, `lastReachable`, `lastUnreachable`) move together through a single `AtomicReference<State>`, so each transition is one linearisation point; readers take a coherent `Snapshot` (used by `BackendsResource`).
- **Access-log timing** (`ProxyRequestsManager`): log on the returned publisher's terminal (`doFinally`) instead of at assembly, so status, backend timing and size are populated.

Includes a test that polls `PENDING_REQUESTS_GAUGE` until the terminal callback fires, removing a race in the prior assertion.

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/api/BackendsResource.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/api/BackendsResource.java
@@ -96,9 +96,10 @@ public class BackendsResource {
             }
             BackendHealthStatus bhs = backendsSnapshot.get(key);
             if (bhs != null) {
-                bean.available = bhs.getStatus() != BackendHealthStatus.Status.DOWN;
-                bean.reportedAsUnreachable = bhs.getStatus() == BackendHealthStatus.Status.DOWN;
-                bean.reportedAsUnreachableTs = bhs.getUnreachableSince();
+                final BackendHealthStatus.Snapshot snap = bhs.snapshot();
+                bean.available = snap.status() != BackendHealthStatus.Status.DOWN;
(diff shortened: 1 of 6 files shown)


NiccoMlt:
@coderabbitai was the nitpick correctly addressed?

dmercuriali (approved):

PR #606, opened 2026-06-17
Fix the Maven release workflow and harden CI

NiccoMlt:
Repairs the `Maven release` workflow so a release runs end-to-end (it previously failed at the deploy step) and cleans up how the release authenticates and attributes its commits.

### How the release authenticates (the core change)
The release needs two different credentials, and one token can't cover both:

- **Git pushes** to the protected `master` / `release/*` branches use a **GitHub App token** — it can bypass branch protection and attributes the release commits to a bot identity.
- **The Maven deploy and the GitHub Release** use the Actions **`GITHUB_TOKEN`**. This is required: the GitHub Packages Maven registry rejects GitHub App tokens, so only a PAT or `GITHUB_TOKEN` authenticates there. Published artifacts and the release show as `github-actions[bot]`.

### Other release fixes
- **`setup-java` credentials** — `server-username`/`server-password` take environment-variable *names*, not values; the previous literal values resolved to an empty password and a 401 on deploy.
- **`pom.xml` SCM** switched to the HTTPS URL with `<tag>HEAD`, so `maven-release-plugin` pushes over HTTPS (matching the token auth) instead of SSH.
- **Commit author** set to `diennea-bot`; the shared Git identity + credential-helper setup is extracted into a local composite action (`.github/actions/configure-git`), defined once and reused by both checkouts.

### Also in this PR (CI hygiene)
Pinned all external actions to commit SHAs, bumped `test-reporter` / `upload-artifact` / `action-gh-release` to their Node 24 majors, added a paths filter so PR validation skips non-code changes, and set `persist-credentials: false` on the read-only checkouts.

### DIFF
--- a/.github/actions/configure-git/action.yml
+++ b/.github/actions/configure-git/action.yml
@@ -1,0 +1,16 @@
+name: 'Configure Git'
+description: 'Configure the diennea-bot Git identity and set up the gh credential helper for pushes'
+inputs:
+  token:
+    description: 'GitHub App token used by the gh Git credential helper for pushes'
+    required: true
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+      run: |
+        git config user.name "diennea-bot"
+        git config user.email "172403682+diennea-bot@users.noreply.github.com"
+        gh auth setup-git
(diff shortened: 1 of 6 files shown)


NiccoMlt:
@coderabbitai resume

NiccoMlt:
@coderabbitai resume

Issue #468, opened 2024-04-10
Configuration cleaned too much


NiccoMlt:
When saving, apparently removes always the backslashes `\` until there are none

NiccoMlt:
Apparently, the UI sends it correctly, so the cleanup doesn't happen client-side, but server-side

PR #617, opened 2026-07-21
Fix config save stripping backslashes from property values

### LINKED ISSUE #608: Config save strips backslashes from property values (connection-pool domain regexes break)
### Summary

Saving dynamic configuration silently drops backslashes from property values. Connection-pool `domain` regexes that use `\Q…\E` (or `\.`, `\d`, …) lose their backslashes and stop matching.

### Cause

`ConfigurationStore.toStringConfiguration()` serializes the configuration as raw `key + "=" + value`, with no escaping. On any change, `HttpProxyServer.rewriteConfiguration()` feeds that text back through `ConfigResource.buildStore()` → `java.util.Properties.load(...)`, which treats `\` as an escape and drops it before unrecognized characters (`\Q`→`Q`, `\E`→`E`).

This path runs for the connection-pool REST API (`/api/connectionpools` POST/PUT/DELETE) and for saving from the config editor, and it re-serializes the **whole** configuration — so every backslash-bearing value is corrupted except the one written in that same call.

### Reproduction

A pool configured with `domain=(\Qhost.example\E)` becomes `domain=(Qhost.exampleE)` after any config save, so `Pattern.matches` no longer matches `host.example`.

### Impact

A connection pool can silently stop matching its hostname and fall back to the default pool — no error, no log. Affects any backslash-bearing property (pool domains, request matchers, rewrite rules).

### Affected versions

Present wherever the config editor / connection-pool CRUD API exist (2.0+); reproduced on 2.1.x and `master`.

### Suggested fix

Serialize via `Properties.store(...)` (which escapes backslashes) instead of `key + "=" + value`, and/or avoid the text round-trip in `rewriteConfiguration`. Add a regression test asserting a `\Q…\E` domain survives a config save.

NiccoMlt:
# Problem

Every config save re-serializes the whole configuration as raw `key=value` text and re-parses it with `Properties.load()`, which treats `\` as an escape and silently drops it. 

Thus, any backslash-bearing value is corrupted: a connection-pool domain like `(\Qhost.example\E)` becomes `(Qhost.exampleE)`, the regex stops matching, and traffic silently falls back to the default pool. Both the connection-pool CRUD API and the config editor hit this path.

# Implementation

Make the round-trip symmetric: `toStringConfiguration()` now serializes with `Properties.store()`, which escapes values exactly as `load()` expects. The timestamp comment `store()` prepends is stripped and lines are kept sorted, preserving the deterministic dump format. 

Fixes #608 #468

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/configstore/ConfigurationStore.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/configstore/ConfigurationStore.java
@@ -20,14 +20,17 @@
 package org.carapaceproxy.configstore;
 
 import herddb.utils.BooleanHolder;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.UncheckedIOException;
 import java.security.KeyPair;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Properties;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.function.BiConsumer;
 import java.util.function.BiPredicate;
+import java.util.stream.Collectors;
 import org.carapaceproxy.core.RuntimeServerConfiguration;
 import org.carapaceproxy.server.config.ConfigurationNotValidException;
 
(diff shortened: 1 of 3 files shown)


dmercuriali (approved):
lgtm

PR #634, opened 2026-08-27
Upgrade the cluster stack, and replace the abandoned lz4

### LINKED ISSUE #633: Upgrade the cluster stack, and replace the abandoned lz4
## Summary

Three Dependabot alerts sit on the clustering stack.

Two are on lz4, both high, and bumping cannot fix them: GHSA-vqf4-7m7x-wgfc, out of bounds memory operations, and GHSA-cmp6-m4wj-q63q, an information leak in the safe decompressor. We ship `net.jpountz.lz4:lz4` **1.3.0**, declared by `herddb-utils`, and that coordinate has had no release since 2014, so neither advisory lists a patched version for it. HerdDB still declares it today, so the only way out is the maintained fork, `at.yawk.lz4:lz4-java`, which carries both fixes as of 1.10.1.

The third is GHSA-gxq5-79m2-gvvq, improper certificate validation in BookKeeper, fixed in 4.14.6. We ship **4.14.3**, the version HerdDB 0.26.1 declares.

That last one hid behind a pin that describes something nobody runs: `libs.bookkeeper` says **4.14.8**, but it governs only `bookkeeper-stats-api` and `prometheus-metrics-provider`. The rest of the family arrives from HerdDB, so the project runs two BookKeeper releases at once, and `dependencyConvergence` cannot see it because it compares versions of the same artifact, not of a family.

## Proposal

- swap `net.jpountz.lz4:lz4` for `at.yawk.lz4:lz4-java`, excluding the old coordinate where HerdDB brings it in
- HerdDB **0.26.1 → 0.28.0**, which declares BookKeeper 4.16.3 and closes GHSA-gxq5-79m2-gvvq
- point `libs.bookkeeper` at what HerdDB actually pulls, and let the enforcer require a single version across the BookKeeper and the ZooKeeper families, so a split cannot go unnoticed again
- Curator **5.2.0 → 5.9.0**, the newest release; HerdDB declares 5.4.0 and nothing else constrains it

> [!NOTE]
> 0.28.0 rather than 0.29.0, which is the latest: 0.29.0 moves to ZooKeeper 3.9, whose client enables FIPS mode by default and then refuses DIGEST-MD5. That is the mechanism the cluster ACLs authenticate with, so `ZookKeeperACLTest` fails with `AuthFailed`. Moving to the 3.9 line needs its own issue and a decision on the SASL mechanism.

ZooKeeper stays on 3.8.6, the newest of its line and free of open advisories.

NiccoMlt:
Closes #633.

Three Dependabot alerts on the clustering stack, none fixable with a plain bump:

- `net.jpountz.lz4:lz4` (GHSA-vqf4-7m7x-wgfc, GHSA-cmp6-m4wj-q63q): no release since 2014, no patched version exists. Replaced with the maintained fork `at.yawk.lz4:lz4-java`.
- BookKeeper GHSA-gxq5-79m2-gvvq: fixed by HerdDB 0.28.0, which brings BookKeeper 4.16.3.

Before the upgrade, `libs.bookkeeper` pointed at 4.14.8 but only governed `bookkeeper-stats-api` and `prometheus-metrics-provider`; the rest of the family came from HerdDB at 4.14.3, so two BookKeeper releases coexisted. `dependencyConvergence` compares versions of a single artifact, not of a family, so it cannot flag this. The first two commits align the pin with the version HerdDB pulls and add a `requireSameVersions` enforcer rule for the BookKeeper and ZooKeeper families.

Also in the PR: Curator 5.2.0 to 5.9.0, and the group membership handler moves off the deprecated `PathChildrenCache` onto `CuratorCache`.

Adjustments that BookKeeper 4.16 required:

- It ships io_uring natives, whose presence alone makes Reactor Netty pick io_uring channels over our epoll event loops; every listener bind fails on Linux. Natives excluded, classes kept (BookKeeper references them unconditionally).
- Its embedded server needs commons-cli, which HerdDB declares with test scope; now declared directly.
- It pulls log4j, bridged through SLF4J instead.

One behavioral note on `CuratorCache`: it replays pre-existing nodes on start, unlike the old `BUILD_INITIAL_CACHE`. Without `afterInitialized()` a restart would re-fire a stale configuration-change event and reload configuration for nothing. `watchEvent` also waits for the initial population, so events fired right after registration cannot fall into the startup window. A restart test covers both.

## Why HerdDB 0.28.0 and not 0.29.0

0.29.0 moves to ZooKeeper 3.9, whose client enables FIPS mode by default, and FIPS forbids DIGEST-MD5, the mechanism our cluster ACLs authenticate with:

```
WARN  ZooKeeperSaslClient - Client will not use DIGEST-MD5 as SASL mechanism, because FIPS mode is enabled.
ERROR ClientCnxn - SASL authentication with Zookeeper Quorum member failed.
javax.security.sasl.SaslException: saslClient failed to initialize properly: it's null.
```

Any deployment using `zookeeper.acl` with SASL digest credentials would break the same way `ZookKeeperACLTest` does. ZooKeeper stays on 3.8.6, newest of its line, no open advisories. The 3.9 move needs its own issue and a SASL decision.

Six commits, merge with rebase, not squash.

### DIFF
--- a/carapace-server/pom.xml
+++ b/carapace-server/pom.xml
@@ -158,7 +158,34 @@
                     <groupId>net.jpountz.lz4</groupId>
                     <artifactId>lz4</artifactId>
                 </exclusion>
+                <!-- BK 4.16 pulls log4j; project uses Logback via SLF4J -->
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-core</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-slf4j-impl</artifactId>
+                </exclusion>
+                <!-- io_uring natives make Reactor Netty pick io_uring over our epoll event loops -->
+                <exclusion>
+                    <groupId>io.netty.incubator</groupId>
+                    <artifactId>netty-incubator-transport-native-io_uring</artifactId>
+                </exclusion>
             </exclusions>
+        </dependency>
+        <dependency>
+            <!-- BK references these classes unconditionally; only the natives are unwanted -->
+            <groupId>io.netty.incubator</groupId>
+            <artifactId>netty-incubator-transport-classes-io_uring</artifactId>
+            <version>0.0.21.Final</version>
+        </dependency>
+        <dependency>
+            <!-- needed by the BK embedded server; HerdDB test-scopes it away -->
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
+            <version>1.11.0</version>
+            <scope>runtime</scope>
         </dependency>
         <dependency>
             <groupId>at.yawk.lz4</groupId>
--- a/carapace-server/pom.xml
+++ b/carapace-server/pom.xml
@@ -153,7 +153,17 @@
                     <artifactId>commons-logging</artifactId>
                     <groupId>commons-logging</groupId>
                 </exclusion>
+                <!-- Abandoned since 2014, replaced by at.yawk.lz4:lz4-java -->
+                <exclusion>
+                    <groupId>net.jpountz.lz4</groupId>
+                    <artifactId>lz4</artifactId>
+                </exclusion>
             </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>at.yawk.lz4</groupId>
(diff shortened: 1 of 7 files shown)


NiccoMlt:
@coderabbitai resume

NiccoMlt:
@coderabbitai resume

Issue #536, opened 2025-05-26
Acme certificate > Support custom acme


hamadodene:
Starting in 2029, TLS certificate lifetimes will be reduced to just 47 days.
More details: https://www.digicert.com/blog/tls-certificate-lifetimes-will-officially-reduce-to-47-days

Many providers already support ACME (e.g., DigiCert), so we need to implement support for multiple ACME providers.

For example, we could allow configuration like this:

```
acme.1.name=digicert
acme.1.url=https://acme.example.com/something
acme.1.apikey=your-api-key

acme.2.name=mycustom_provider
acme.2.url=https://acme.example.com/something
acme.2.apikey=your-api-key
```

Then, when generating a certificate, users could specify the ACME provider:

```
certificate.1.hostname=myhostname.example.com
certificate.1.mode=acme
certificate.1.provider=mycustom_provider

```

By default, if no provider is specified, we can fall back to Let's Encrypt (LE) --> LE must be default provider.

Note:
The API and UI will also need to be updated. However, we can create follow-up issues to track some of these changes separately.

PR #653, opened 2026-09-07
Harden the certificates lifecycle: idempotent challenge tokens, counted failures, quiet mutex

NiccoMlt:
Four fixes to `DynamicCertificatesManager` and its cluster mutex, found while triaging a 2.3.0 rollout with about a thousand ACME certificates. Each is its own commit.

**Challenge tokens are now upserted.** `saveAcmeChallengeToken` did a plain `INSERT` into `acme_challenge_tokens`. The CA reuses a pending authorization, and with it the same token, when the same domain is ordered again within a few days, so the second order hit the primary key with `DuplicatePrimaryKeyException`. The resulting `ConfigurationStoreException` is a `RuntimeException`, which the lifecycle loop did not catch: the run died there and every domain after it in alphabetical order was never processed again. HerdDB supports `UPSERT INTO` natively, so the statement just changes keyword. The loop now also catches `RuntimeException` (a store failure is logged and retried at the next cycle, not counted against the certificate), so one broken domain can no longer take the others down with it. `master` has the same `INSERT` and needs the first half of this commit.

**Non-transient failures are counted on the certificate.** Partial backport of issue #536 → PR #620: `AcmeFailureClassifier` and its test are taken verbatim, and the lifecycle catch now moves the certificate to `REQUEST_FAILED` and increments its attempts unless the failure is transient (network error, rate limit, CA outage, bare 5xx, store failure). Before this a CA rejection or a stale order was logged and retried identically every period, forever, without ever surfacing in the UI. After `dynamiccertificatesmanager.errors.maxattempts` the domain stays parked until a reset. This is a behaviour change worth knowing about when reviewing. Not needed on `master`.

**The cluster mutex is released only when it was acquired.** `executeInMutex` released in `finally` even after a failed `acquire`, so every node that is not the current leader logged `IllegalMonitorStateException` at ERROR on each period. A failing runnable was also logged as a lock acquisition failure; it now has its own message.

**Certificates parked past the attempts limit are left alone.** A `DOMAIN_UNREACHABLE` or `REQUEST_FAILED` certificate beyond the limit was still saved on every run, and that save flagged the run as changed: a single dead domain in the configuration made every node reload all certificates from the database each period, with the full `RELOADED` log for each of them.

Each commit comes with a regression test that fails on `release/2.3` as it is. The mutex test captures the handler's log with a Logback `ListAppender`, since the bug only shows up as a logged exception.

The last commit is the development version bump of `release/2.3` to `2.3.1-SNAPSHOT`, cherry-picked from PR #652 so that this branch builds as the next patch release.

Targets `release/2.3`. Forward-port to `master`: the `UPSERT` from the first commit, and the third and fourth commits as clean cherry-picks.

### DIFF
--- a/carapace-server/pom.xml
+++ b/carapace-server/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.carapaceproxy</groupId>
         <artifactId>carapace-parent</artifactId>
-        <version>2.4.1-SNAPSHOT</version>
+        <version>2.3.1-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
--- a/carapace-server/src/main/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandler.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/cluster/impl/ZooKeeperGroupMembershipHandler.java
@@ -290,20 +290,26 @@ public class ZooKeeperGroupMembershipHandler implements GroupMembershipHandler, 
         InterProcessMutex mutex = mutexes.computeIfAbsent(mutexId, (mId) -> {
             return new InterProcessMutex(client, "/proxy/mutex/" + mutexId);
         });
+        boolean acquired = false;
         try {
-            boolean acquired = mutex.acquire(timeout, TimeUnit.SECONDS);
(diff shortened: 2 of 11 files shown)


NiccoMlt:
Let's not rush fixes against release branches.

I'm closing this, we will follow the standard HF procedure with the approved changes.

Issue #654, opened 2026-09-13
Allow configuring the HerdDB tablespace used for the configuration store


NiccoMlt:
Carapace always keeps its configuration tables in HerdDB's default tablespace, `herd`. The name is not configurable: `HerdDBConfigurationStore` never sets a schema, so the datasource stays on `TableSpace.DEFAULT`, and in cluster mode the startup wait is a `"herd"` literal.

This makes restoring a configuration dump awkward. HerdDB restores into a new tablespace, and there is no way to point Carapace at it, so you either keep the proxy down for the whole restore or work on the live tablespace while it serves traffic. Being able to name the tablespace, defaulting to the current one, would be enough.

PR #655, opened 2026-09-13
Allow configuring the HerdDB tablespace used for the configuration store (2.1)

NiccoMlt:
Backport of issue #654 → PR #657 to `release/2.1`. The 2.3 counterpart is PR #658.

The first commit is the one deployed during the September incident; the two that follow align this branch with master, so the property behaves identically on all three lines.

## Problem

The configuration tables always live in HerdDB's default tablespace, `herd`, and there is currently no way to point Carapace at a different one. This makes restoring a dump awkward: HerdDB restores into a new tablespace, so the proxy must either stay down for the duration of the restore, or keep operating on the tablespace that is serving traffic.

## Change

This pull request introduces a `db.tablespace` property. When it names a tablespace other than the default, the configuration store creates it at boot if it does not exist, waits for it to become available, and points the datasource at it. In cluster mode the tablespace is created with `expectedreplicacount` taken from `replication.factor`; a standalone node always uses 1.

The default is unchanged, so existing deployments remain on `herd` and never enter the new code path.

## Restore procedure

The order matters: HerdDB creates the destination tablespace as part of the restore and fails if the name is already registered. `db.tablespace` must therefore be set only after the restore has completed. If it is set beforehand, the proxy registers the name at boot and the restore fails.

The same value must be configured on every node of a cluster.

Refs #654

### DIFF
--- a/carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
+++ b/carapace-server/src/main/java/org/carapaceproxy/configstore/HerdDBConfigurationStore.java
@@ -27,7 +27,11 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import herddb.client.ClientConfiguration;
+import herddb.client.ClientSideMetadataProviderException;
+import herddb.client.HDBException;
+import herddb.jdbc.HerdDBConnection;
 import herddb.jdbc.HerdDBEmbeddedDataSource;
+import herddb.model.TableSpace;
 import herddb.security.SimpleSingleUserManager;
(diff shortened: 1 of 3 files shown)


NiccoMlt:
@coderabbitai approve

.github/workflows/pr-validation.yml
```diff
@@ -51,8 +51,6 @@ jobs:
           cache: 'maven'
       - name: 'Build and test with Maven'
         run: ./mvnw -B package --file pom.xml
-      - name: 'Submit Dependency Snapshot'
-        uses: advanced-security/maven-dependency-submission-action@v4
```
dmercuriali: why are we removing this in this pr?

NiccoMlt: Because otherwise the CI will continue failing for PRs against this release branch.

It is a backport of a fix from `master` and 2.3

Issue #656, opened 2026-09-13
Throttle how many certificates DynamicCertificatesManager takes to the CA in a single run


NiccoMlt:
Every run of `DynamicCertificatesManager` advances each non-manual certificate by one step, and six of the nine lifecycle states reach the ACME server: the new order with its authorizations, the DNS challenge trigger, the challenge poll, the finalization, the order poll. Nothing bounds that loop, so the work a run generates grows with the number of domains; past a thousand of them a single run takes the better part of an hour, and for all that time the account keeps opening orders at whatever rate the loop manages, which is how an account gets close to the limits the CA enforces on it.

A ceiling on how many certificates may take one of those steps in a run, with the rest left to the next run, is enough to bound it, and it should be configurable so that an installation with few domains does not pay for one with thousands.

The order in which the slots are handed out matters as well: domains are taken alphabetically and each keeps its slot until its certificate is issued or fails, so with more domains than slots the tail of the list must not end up waiting for the head indefinitely.

PR #660, opened 2026-09-18
Throttle how many certificates DynamicCertificatesManager takes to the CA in a single run (2.3)

NiccoMlt:
Aligns `release/2.3` with issue #656 → PR #659.

## Problem

`dynamiccertificatesmanager.ratelimit` has been on this line since #651 was merged, and it is not in any released version: the only tag of the line, `v2.3.0`, predates it. That first pass was written quickly and left `DNS_CHALLENGE_WAIT` out of the count, although that step logs into the account and triggers the challenge, so it reaches the CA like the states that are counted. A certificate waiting for DNS propagation therefore talks to the CA outside the ceiling.

Two more things differ from what #659 does on `master`: a negative value is accepted and silently behaves as no limit, and the slots are always handed out from the head of the alphabet, so with fewer slots than domains the tail waits for the head to finish.

## Change

`isAcmeStep` now counts `DNS_CHALLENGE_WAIT` and lists all nine states, so a new one breaks the build instead of being absorbed by a `default` arm. The rest of the classification is the one this line already had:

| State | Takes a slot | Why |
|---|---|---|
| `WAITING` | yes | new order, authorizations, challenge trigger |
| `DOMAIN_UNREACHABLE` | while under `dynamiccertificatesmanager.errors.maxattempts` | same path; beyond the attempts the step does nothing |
| `DNS_CHALLENGE_WAIT` | yes | account login and challenge trigger |
| `VERIFYING` | yes | account login and one challenge poll per name |
| `VERIFIED` | yes | order binding and finalization |
| `ORDERING` | yes | order binding, poll and certificate download |
| `REQUEST_FAILED`, `AVAILABLE`, `EXPIRED` | no | local transitions |

A negative value is rejected at configuration time, next to the check `healthmanager.connecttimeout` already has. The per-domain `DEBUG` line is replaced by one `INFO` line at the end of a run that hit the ceiling, with how many certificates it postponed. The name of the property, its default of 300 and the meaning of `0` do not change, so an installation upgrading from this line keeps its configuration.

The ceiling is per run, and every node runs the lifecycle on its own schedule: the mutex serialises those runs, it does not elect one node. A cluster of N nodes can therefore spend N ceilings in a period, unless a run overruns `dynamiccertificatesmanager.period` and the others time out on the lock.

Each run now restarts from the domain that follows the last one served. The cursor lives in memory and each node keeps its own, since the lock changes hands run after run: a node that has just started begins from the head, then walks the whole list like the others.

The lifecycle step is extracted into `advance(domain, cert)`, which returns whether the certificate has to be saved: `certificatesLifecycle` is left with the iteration, the ceiling and the save, and the two files read the same way on both lines.

## Tests

`testRateLimitPerRun` is replaced by the three of #659: the ceiling itself, the classification, and the rotation. A fourth, in a new `RuntimeServerConfigurationTest`, covers the rejected value.

Fixing the order in which the slots are handed out also exposed a latent problem in the suite: `carapace.acme.dnschallengereachabilitycheck.limit` is read into a `static final` field but was set from inside two test methods, so it only worked while those happened to run first. It is now set in a static initializer of the test class, as on `master`.

Refs #656


NiccoMlt:
@coderabbitai rate limit

NiccoMlt:
@coderabbitai review

dmercuriali (approved):

